PR #1, opened 2019-07-22
msgraph-sdk-raptor design spec

titusgicheru:
**Design spec for the E2E functional test**

- The spec has two different strategies on how to implement the above.
- The intention is to have a discussion and iterate if need be.


msgraph-sdk-raptor-design-spec.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Requirements
+- Enable all the skipped test within the existing client libraries.
+- Setup **demo tenant** account
+- **Data Integrity** - Maintain demo tenant data state – basically implement this continuously across SDKs without degrading the test data.
+- **Authentication Flows** 
+    - Implement multiple authentication flows with an intent to test both delegated and application permissions.
+
+## Strategy One
+**Enable all the current disabled functional tests and implement ability to hit the service using a demo tenant as opposed to using mocks**
```
MIchaelMainer: Enable on the non destructive functional tests

MIchaelMainer: Enable only the non-destructive e2e tests.

msgraph-sdk-raptor-design-spec.md
```diff
@@ -0,0 +1,64 @@
...
+
+### This approach will focus on;
+- Enabling all the disabled integration test on the test projects on the existing libraries. 
+- Hit the service with each workload and expecting.
+
+### Drawbacks
+- The whole process of enabling the test is manual.
+- Introduction of new features within the API means adding the equivalent tests manually
+- Breaking changes also means changing the test manually
```
MIchaelMainer: This should happen only rarely - this shouldn't be a concern.

msgraph-sdk-raptor-design-spec.md
```diff
@@ -0,0 +1,64 @@
...
+- Service interruptions means the test will fail
```
MIchaelMainer: Excellent point. This will happen and we don't want this to block us. Ideally failed e2e tests should be gated warnings. I'm not sure how that would work in a build, but possibly could be used in a release pipeline.

msgraph-sdk-raptor-design-spec.md
```diff
@@ -0,0 +1,64 @@
...
+- Data within the demo tenant will inevitably degrade over time.
+
+## Strategy Two
+**Implement test by compiling generated code snippets**
+
+### The approach will focus on;
+- Compiling generated code snippets as another way for testing contrary to the first approach.
+- Use of a compiler to achieve this. Roslyn seem to be a good option
+- Predetermine execution order given that some requests are dependent on a sequence and verify responses are as expected.
```
MIchaelMainer: This approach is a smoke test for tests and has nothing to do with e2e testing. There will be no errors, only build success and failure. 

I've got a feeling I'm missing something important to understand this PR :)

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -0,0 +1,91 @@
+﻿using Microsoft.CodeAnalysis;
```
MIchaelMainer: Let's compile this from the command line. We should do compilation from CLI so that we use the same mechanism to get the language snippets to the routine that will pass the snippet from the commandline. Then, we just need to make sure that the correct compiler and command line syntax is used for the target language.

MIchaelMainer:
Testing the code snippets will require us have a weekly release of our client library so that it represents the latest v1.0 documented features, including the generated snippets. We don't want to have snippets generated for a feature that has not yet been generated in the client library. That is assuming we validate the snippets weekly and that there are weekly updates to the v1.0 metadata.

msgraph-sdk-raptor-design-spec.md
MIchaelMainer: I know why I feel this way. The scope appears to have moved from test cases in the client library repo to testing the code snippets. We shouldn't conflate these two goals.

PR #30, opened 2020-05-05
Mark known issues as "skipped" instead of "failed"

zengin:
- Added support for identifying known issues and reporting them as `skipped`
  - Added [an issue](https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/490) regarding composable functions on Range() and their associated test cases
  - Added several test cases from v1.0 where HTTP snippets need to be fixed
- Upgraded nuget packages for Graph
- Added `Newtonsoft.Json.Linq` to compilation template to resolve `JObject` type
- Refactored test case data into a class of its own to avoid passing multiple parameters to test cases

#AB4682

### DIFF
--- a/TestsCommon/TestDataGenerator.cs
+++ b/TestsCommon/TestDataGenerator.cs
@@ -2,7 +2,6 @@
 
 using MsGraphSDKSnippetsCompiler.Models;
 using NUnit.Framework;
-using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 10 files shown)


TestsCommon/TestDataGenerator.cs
```diff
@@ -2,14 +2,152 @@
...
+        {
+            return new Dictionary<string, string>()
+            {
+                { "call-transfer-csharp-V1-compiles", GetPropertyNotFoundMessage("InvitationParticipantInfo", "EndpointType") },
+                { "create-educationschool-from-educationroot-csharp-V1-compiles", GetPropertyNotFoundMessage("EducationSchool", "Status") },
+                { "create-item-attachment-from-eventmessage-csharp-V1-compiles", HttpSnippetWrong + ": Item needs to be an OutlookItem object, not a string" },
+                { "create-rangeborder-from-rangeformat-csharp-V1-compiles", FeatureNotSupported },
+                { "create-reference-attachment-with-post-csharp-V1-compiles", GetPropertyNotFoundMessage("ReferenceAttachment", "SourceUrl, ProviderType, Permission and IsFolder") },
+                { "create-tablecolumn-from-table-csharp-V1-compiles", HttpSnippetWrong + "Id should be string not int" },
+                { "get-borders-csharp-V1-compiles", FeatureNotSupported },
```
darrelmiller, line 51: I think this might be the clearest indication that we really should add composable function support into the SDK!  It might be worth getting some usage data on these functions to understand the number of apps that actually use these capabilities.

PR #33, opened 2020-05-07
Split test runs for known issues

zengin:
The purpose of this change is to provide feature owners a web link with all the actionable feedback regarding failures.

Azure DevOps has a few limitations around working with known issues.
1. I have investigated the option of marking known issues as "ignored" or "inconclusive". In these cases Azure DevOps doesn't show the error message, which contains the documentation link and other details about the failure. So we can't provide a web link to interested parties, such as owners that need to fix HTTP snippets.
2. Azure DevOps also has a UI to add analysis for individual failures, but these analyses are per each pipeline run. In other words, they do not stick to the test cases even though the system is able track the history of a test properly for failed/passed states.

With this change, I am splitting test projects into two as known issues and others. I also generated two new pipelines for known issues.

The results now look like the following:
| Title | Failed | Pass Rate |
|--|--|--|
| Beta C# Snippet Compilation Tests | 153 | 87.8% |
| V1 C# Snippet Compilation Tests | 64 | 90.5% |
| Beta C# Snippet Compilation Tests - Known Issues | 40 | 0% |
| V1 C# Snippet Compilation Tests - Known Issues | 51 | 0% |

In this model, the target is to get to 100% for the first two initially, while the last two represent the technical debt.

### DIFF
--- a/azure-pipelines/csharp-beta-compile-tests-known-issues.yml
+++ b/azure-pipelines/csharp-beta-compile-tests-known-issues.yml
@@ -1,0 +1,52 @@
+# Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+
+trigger: none # disable triggers based on commits.
+
+resources:
+ repositories:
+   - repository: microsoft-graph-docs
+     type: github
+     endpoint: microsoftgraph
+     name: microsoftgraph/microsoft-graph-docs
+     ref: master
+
+pool:
+  vmImage: 'windows-2019'
+
+variables:
+  buildConfiguration: 'Release'
+
+steps:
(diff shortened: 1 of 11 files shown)


azure-pipelines/csharp-beta-compile-tests-known-issues.yml
```diff
@@ -0,0 +1,52 @@
...
+resources:
+ repositories:
+   - repository: microsoft-graph-docs
+     type: github
+     endpoint: microsoftgraph
+     name: microsoftgraph/microsoft-graph-docs
+     ref: master
+
+pool:
+  vmImage: 'windows-2019'
```
MIchaelMainer, line 14: nit: use `windows-latest` to save an update 4-6 years from now, same for all other pipelines

PR #38, opened 2020-05-14
Remove passing test from known issues

zengin:
This test started passing after getting the latest Beta update.


darrelmiller:
Tests that fix themselves is what I am all about :-)

PR #48, opened 2020-07-08
Add LinqPad support for quick local analysis

zengin:
- Added logic to auto generate LinqPad query files for compilation failures. Files are dumped into default `My Queries` folder. This makes compilation failures very easy to detect (certainly way faster than copying snippets from test results into a .cs file with correct Nuget references etc).
- Added settings to separate local runs from cloud runs, instead of inferring them from existence of `BUILD_SOURCESDIRECTORY` environment variable.
- Added notes into README.md
- Removed left over sql queries from earlier design

cc: @pixieofhugs


MIchaelMainer:
@zengin We should consider integration with https://github.com/BrianTJackett/graph-test-harness as it has a harness all setup.

README.md
```diff
@@ -1 +1,34 @@
...
+1. Clone this repository
+2. Clone microsoft-graph-docs repository:
+   - `git clone https://github.com/microsoftgraph/microsoft-graph-docs`
+3. Open msgraph-sdk-raptor.sln in Visual Studio
+4. Make sure that the settings are correct for local run in `msgraph-sdk-raptor\msgraph-sdk-raptor-compiler-lib\appsettings.json`
+   1. `"IsLocalRun"=true`
+   2. `"LocalRootGitDirectory"= <where microsoft-graph-docs repo is located>`
+5. Build and run tests
+   - Test count will show as 1 for each project initially because test cases are generated on the fly from a single meta test description.
+6. Local test runs also generate `.linq` files so that they can be analyzed using LinqPad. Default drop location for these files is:
```
MIchaelMainer, line 28: It may make sense to use the harness as a template and inject the failed test cases into the template created by Brian.

zengin:
@MIchaelMainer test harness looks awesome. At this point, this is only generating for the purpose of compilation. But once we move into the phase where we run snippets against actual endpoints, I think integration would be really valuable.

TestsCommon/CSharpTestRunner.cs
```diff
@@ -111,7 +111,70 @@ public static void Run(CsharpTestData testData)
...
         }
+
+        /// <summary>
+        /// Generates .linq file in default My Queries folder so that the results are visible in LinqPad right away
+        /// </summary>
+        /// <param name="testData">test data</param>
+        /// <param name="codeSnippetFormatted">code snippet</param>
+        private static void WriteLinqFile(CsharpTestData testData, string codeSnippetFormatted)
+        {
+            var linqPadQueriesDefaultFolder = Path.Join(
```
MIchaelMainer, line 132: We should validate that this path exists and exit gracefully if it doesn't (perhaps with a message with explanation). Later at Directory.CreateDirectory, it will throw DirectoryNotFoundException if LinqPad isn't installed, or it is installed with a non default query location.

zengin: CreateDirectory creates the full path if it doesn't exist, DirectoryNotFoundException can happen if the drive is invalid, but `USERPROFILE` makes sure that the drive exists.

We can have a setting to prevent creating LinqPad files if the developer doesn't have LinqPad installed, because the current code will always generate these outputs for no use case.

BrianTJackett:
Let me know if anything I can assist with in the future. Haven't touched the test harness repo lately but I use it almost daily for testing purposes.

PR #66, opened 2020-08-18
Add more known issues in Beta

zengin:
- Spent some time with @MIchaelMainer to analyze some of the remaining Beta issues.
- Initial pass on issues are completed both for V1 and Beta. I will make another pass on the known issues since I understand the overall system better now.
- I also removed some of the summary tags as most of the names/string constants were self explanatory. Instead I grouped them under `#region`s.

With this change, Beta/V1 tests should all pass 🤞, except for the ones marked as known.

### DIFF
--- a/msgraph-sdk-raptor-compiler-lib/msgraph-sdk-raptor-compiler-lib.csproj
+++ b/msgraph-sdk-raptor-compiler-lib/msgraph-sdk-raptor-compiler-lib.csproj
@@ -12,7 +12,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="3.1.7" />
     <PackageReference Include="Microsoft.Graph" Version="3.9.0" />
     <PackageReference Include="Microsoft.Graph.Auth" Version="1.0.0-preview.5" />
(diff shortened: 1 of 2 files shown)


msgraph-sdk-raptor-compiler-lib/msgraph-sdk-raptor-compiler-lib.csproj
```diff
@@ -12,7 +12,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="3.1.7" />
     <PackageReference Include="Microsoft.Graph" Version="3.9.0" />
     <PackageReference Include="Microsoft.Graph.Auth" Version="1.0.0-preview.5" />
-    <PackageReference Include="Microsoft.Graph.Beta" Version="0.20.0-preview" />
+    <PackageReference Include="Microsoft.Graph.Beta" Version="0.22.0-preview" />
```
zengin, line 15: Beta package just got updated.
```suggestion
    <PackageReference Include="Microsoft.Graph.Beta" Version="0.23.0-preview" />
```

zengin:
4 more snippets with problems added to docs, so Beta is still not fully passing. I will add those as a separate PR.

PR #96, opened 2020-10-30
End to end Raptor Pipelines for Staging Beta and Staging V1 (C# Compilation Tests)

zengin:
- `beta-end-to-end.yml` and `v1-end-to-end.yml` are the main pipelines. They set metadata URL and version information from parameters. They then include the following templates:
  - `variables.yml`: contains shared variables, version and URL parameters are used to set variables too.
  - `checkouts.yml`: checkout steps
  - `snippet-generation.yml`:
    - Builds snippet generator from `microsoft-graph-explorer-api`
    - Builds and runs `apidoctor` consuming snippet generator command line tool
  - `sdk-geneation.yml`:
    - Transforms given metadata and builds a new `Microsoft.Graph.dll`
  - `regular-tests.yml`:
    - Transforms `Test.runsettings` file to refer to the Microsoft.Graph.dll generated in previous step
    - Runs Raptor tests that are supposed to be passing (i.e. no known issues)
  - `known-failure-tests.yml`
    - This step is added only if the parameter for running known failures is true
    - Similar to the step above, only it runs the known failure tests (This could be useful if staging has any fixes for known issues)
  - `generate-type-summary.yml`
    - Generates a type summary for the built Microsoft.Graph.dll for debugging purposes. It is published as an artifact at the end.

Other refactorings and changes to make the pipelines work:
- Generalized Raptor project to consume arbitrary `Microsoft.Graph.dll`s that are generated on the fly. The path to generated dll is passed in a `Test.runsettings` file. And it gets loaded from that path instead of the Nuget reference.
- Created a `transformSettings.ps1` script to change test settings within pipelines
- Created a new `RunSettings` object to pass around test settings
- Added `arbitraryDll` into test names to distinguish them from other tests (Even though it is not perfect, Azure DevOps keeps a history based on test name).

### Links
[Initial dev design document](https://github.com/microsoftgraph/msgraph-sdk-design/blob/master/raptor/pipeline.md)
[Beta Run](https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=30577&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9)
[V1 Run](https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=30587&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=2b20c7d0-7587-5b0f-aadc-b456e6f0b807)

### Work Items
[AB#6095](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6095)
[AB#6096](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6096)
[AB#6097](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6097)
[AB#6098](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6098)
[AB#6099](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6099)
[AB#6100](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6100)
[AB#6101](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6101)

cc: @darrelmiller @pixieofhugs


zengin:
Please ignore the Java checks, they are work in progress.

azure-pipelines/e2e-templates/sdk-generation.yml
```diff
@@ -0,0 +1,68 @@
...
+
+- task: VSBuild@1
+  displayName: 'Build Typewriter'
+  inputs:
+    solution: '$(typewriterSolution)'
+    platform: 'Any CPU'
+    configuration: '$(buildConfiguration)'
+
+# TODO support metadata path
+- task: PowerShell@2
```
baywet, line 18: ideally for this one we should be able to reuse some of the pipeline definition in the repo or at least the script to avoid duplicating maintenance

baywet (commented):
this is great! two questions:
- Are you planning to do the equivalent for Java?
- Shouldn't the language steps (generation + running the tests) run as a matrix? this way if we add java or other languages, it'll run in parallels for each language instead of sequentially

zengin:
> this is great! two questions:
> 
> * Are you planning to do the equivalent for Java?
> * Shouldn't the language steps (generation + running the tests) run as a matrix? this way if we add java or other languages, it'll run in parallels for each language instead of sequentially

Makes sense as other languages are part of the plan, but the original plan was: C# compilation -> C# execution with demo tenant -> introducing other languages.

I think that compilation surfaces certain set of errors, but execution will have more added value as we are not doing the execution at all at the moment. Hence from the perspective of what I should spend my effort on, the prioritization of snippet execution makes more sense to me than introducing multi language support.

That said, you have done a great job in generalizing PROD runs to multiple languages. If you think we should start running your tests for staging as well, be my guest to modify the pipeline. I think what you pointed out should be possible to do when we introduce other languages but if you think that current state of the pipeline makes introducing new languages really hard, i.e. it requires a complete overhaul, please let me know.

azure-pipelines/e2e-templates/sdk-generation.yml
baywet: would you mind getting the PowerShell out of the YAML into it's own ps1 file as a first step towards reusability?
It makes it easier to read (syntax etc) and it'll point people reading/modifying the pipelines to do the same thing, setting good precedents.

baywet:
prioritizing execution against a demo tenant vs compiling multiple languages makes sense. I think integration testing is where we hurt the most at this point. Thanks for taking the time to explain. I'm working to get the java pipeline cleaned up to a baseline so we can start relying on it.
As per your comment about the ease of integrating another language, I think it's fairly straightforward, and it'll be even easier for the next language as now the infrastructure is a bit more generic and as we have a template for non-roselyn languages.

TestsCommon/RunSettings.cs
```diff
@@ -0,0 +1,55 @@
...
+                "JAVA" => Languages.Java,
+                "JAVASCRIPT" => Languages.JavaScript,
+                "JS" => Languages.JavaScript,
+                "OBJC" => Languages.ObjC,
+                "OBJECTIVEC" => Languages.ObjC,
+                "OBJECTIVE-C" => Languages.ObjC,
+                _ => Languages.CSharp
+            };
+
+            if (!File.Exists(dllPath))
```
MIchaelMainer, line 45: Why does `dllPath` exist for languages other than C#?

scripts/generateModels.ps1
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
```
MIchaelMainer, line 1: nit: since this is specific for .NET, I suggest renaming to generateCSharpModels.ps1.

MIchaelMainer (approved):
I like the organization of these pipelines.

TestsCommon/RunSettings.cs
zengin: Good catch. Thanks!
This setting file is only used by C# at the moment. So this part was written before the introduction of Java compilation yesterday.

I had to have a language option for downstream compatibility with changes merged in yesterday, so added the parsing logic above after the merge. I will add a check here that the language should be C#.

MIchaelMainer (approved):

scripts/generateModels.ps1
zengin: changed

nikithauc:
> I like the organization of these pipelines.

+1

azure-pipelines/e2e-templates/known-failure-tests.yml
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+# Steps to run known failure tests
```
nikithauc, line 2: It would be nice to specify the purpose of having a known failure tests same as what you have mentioned in the description.

nikithauc (approved):

PR #106, opened 2020-11-10
feature/java preview

baywet:
changes in this pull request:
- adds randomization of java compile folder (to avoid conflicts in case of parrallels execution)
- adds runsettings  configuration files (with parallelization of work for java)
- adds defensive programming for the runsettings parser to avoid taking into account default/empty values
- adds the ability for the java test runner to run tests against a preview version of the SDK
- adds pipelines to run tests against a preview version of the SDK (using resources and steps that build the SDK first)


TestsCommon/RunSettings.cs
```diff
@@ -29,27 +29,32 @@ public RunSettings(TestParameters parameters)
...
+                    "OBJECTIVEC" => Languages.ObjC,
+                    "OBJECTIVE-C" => Languages.ObjC,
+                    _ => Languages.CSharp
+                };
+
+            if (!string.IsNullOrEmpty(dllPath) && !dllPath.Contains(dashdash))
             {
-                throw new ArgumentException("File specified with DllPath in Test.runsettings doesn't exist!");
+                DllPath = dllPath;
+                if ((Language == Languages.CSharp || Language == Languages.Java) && !File.Exists(dllPath) && !Directory.Exists(dllPath)) // java uses a directory
```
zengin, line 51: Please use a different parameter for Java. Let's keep `dllPath` for Microsoft.Graph.dll.

JavaV1KnownFailureTests/Test.runsettings
```diff
@@ -0,0 +1,12 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RunSettings>
+  <RunConfiguration>
+    <MaxCpuCount>4</MaxCpuCount>
+  </RunConfiguration>
+  <TestRunParameters>
+    <Parameter name="Version" value="---VersionPlaceholder---" />
+    <Parameter name="DllPath" value="C:/sources/github" />
+    <Parameter name="KnownFailuresRequested" value="---KnownFailuresRequestedPlaceholder---" />
+    <Parameter name="Language" value="---LanguagePlaceHolder---" />
```
zengin, line 10: Don't we know these values beforehand for these specific projects, i.e. Version = V1, KnownFailuresRequested=true, Language=Java.

zengin: Maybe we can hardcode these and skip the transform steps.

zengin (changes requested):

JavaV1KnownFailureTests/Test.runsettings
baywet: my idea here was that I didn't want to create another 2 test projects for the preview version. So any java project can be used to test against a preview version.

TestsCommon/RunSettings.cs
baywet: fixed

baywet:
@zengin this is ready for final review. I requeued the existing pipelines to double check I didn't break anything. And here is a run with the Java preview one. (with correct branches for resources)
https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=31836&view=results

zengin (approved):

PR #107, opened 2020-11-11
Remove passing tests from known issues

zengin:
- Issue with snippet generation is fixed: https://github.com/microsoftgraph/microsoft-graph-explorer-api/issues/297
  - Removed the corresponding issues from known issues.
- Removed unreferenced fields/constants.

### DIFF
--- a/TestsCommon/TestDataGenerator.cs
+++ b/TestsCommon/TestDataGenerator.cs
@@ -539,6 +539,8 @@ namespace TestsCommon
 
                 {$"upload-via-put-id-java-{version}-compiles", new KnownIssue(SnippetGeneration, "Missing support for content: https://github.com/microsoftgraph/microsoft-graph-explorer-api/issues/371") },
 
+                {"create-printer-java-Beta-compiles", new KnownIssue(SnippetGeneration, "Parameters with null values are not accounted for as action parameters") },
+
                 {"get-group-java-Beta-compiles", new KnownIssue(TestGeneration, "Imports need to be deduplicated for namespaces") },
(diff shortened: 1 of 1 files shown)


TestsCommon/TestDataGenerator.cs
```diff
@@ -247,7 +233,6 @@ private static string GetTypeNotDefinedMessage(string typeName)
                 { $"update-workforceintegration-{lng}-V1-compiles", new KnownIssue(HTTP, HttpSnippetWrong + ": workforceintegration id is needed in the url.") },
                 { $"worksheet-range-{lng}-Beta-compiles", new KnownIssue(HTTPMethodWrong, GetMethodWrongMessage(POST, GET)) },
                 { $"get-channel-messages-delta-1-{lng}-V1-compiles", new KnownIssue(Metadata, "Delta function is not declared") },
-                { $"create-printer-{lng}-Beta-compiles", new KnownIssue(SnippetGeneration, SnippetGenerationTypeHint) },
```
baywet, line 250: QQ for this one, it's not erroring anymore in java either?

baywet (approved):

TestsCommon/TestDataGenerator.cs
zengin: It fails in Java with a similar root cause, but it is not exactly the same, so we shouldn't have bucketed this one with the C# issue:
https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=31883&view=ms.vss-test-web.build-test-results-tab&runId=1006110&resultId=101210&paneView=debug

C# was missing the type hint (var param = null), but Java seems to be missing the parameter altogether.

I will add it back to the Java issues list with the root cause.

baywet (approved):

PR #115, opened 2020-12-02
Add default .NET .editorconfig and apply the suggestions for existing code

zengin:


andrueastman (approved):

baywet (commented):
The only thing I find with this configuration is that forcibly adding if/else curly braces for conditional blocks, even when they are single line, makes things "heavier". But I get that's a preference thing.

PR #130, opened 2020-12-09
Remove reviewer list from dependabot YAML, rely on CODEOWNERS file

zengin:


zengin:
@baywet, CODEOWNERS worked with this manual PR creation as well. I didn't add anyone explicitly to this PR.

baywet (approved):

baywet:
@zengin perfect, thanks for confirming. Get ready for a barrage of pull requests tomorrow adding it to different repos from my side :D

zengin:
@baywet I would be more than happy to review those to save me time in the long run. Searching reviewers one by one was so terrible that I was planning to write a bookmarklet in JS to automate that. I don't know why I didn't do a web search on that though, I am pretty sure CODEOWNERS would have showed up if I did.

baywet:
@zengin hey don't feel bad about it. Same situation here, I thought about trying to automate it as well and only discovered the feature at GitHub universe today... Good thing it was here only for 3 years already...

PR #131, opened 2020-12-09
Known issues

zengin:
- Remove isFavoriteByDefault from V1 known issues as the property is now in V1.
- Add the new failure with PutAsync due to SDK's missing feature.

### DIFF
--- a/azure-pipelines/java-v1-compile-tests-known-issues.yml
+++ b/azure-pipelines/java-v1-compile-tests-known-issues.yml
@@ -25,7 +25,7 @@ steps:
   parameters:
     projectFileName: JavaV1KnownFailureTests
     javaPreviewLibPath: ''
-    javaLibVersion: $serviceLibraryVersion
+    javaLibVersion: $(serviceLibraryVersion)
 - template: compile-run-tests-template.yml
   parameters:
     projectFileName: JavaV1KnownFailureTests
--- a/azure-pipelines/java-v1-compile-tests-known-issues.yml
+++ b/azure-pipelines/java-v1-compile-tests-known-issues.yml
@@ -20,10 +20,12 @@ variables:
 steps:
 - template: java-templates/java-prepare.yml
 - template: common-templates/checkout.yml
+- template: java-templates/java-set-v1-version.yml
 - template: java-templates/java-replace-local-path.yml
   parameters:
     projectFileName: JavaV1KnownFailureTests
     javaPreviewLibPath: ''
+    javaLibVersion: $serviceLibraryVersion
 - template: compile-run-tests-template.yml
   parameters:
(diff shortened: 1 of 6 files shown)


MIchaelMainer (approved):

baywet (approved):

baywet:
@zengin QQ: shouldn't the test be moved to the java language instead of being removed completely since they [are still failing](https://github.com/microsoftgraph/msgraph-sdk-raptor/pull/131/checks?check_run_id=1522275060)?

baywet:
@zengin since the tests [are still failing](https://github.com/microsoftgraph/msgraph-sdk-raptor/pull/131/checks?check_run_id=1522275060) for java, shouldn't we move those exceptions in the java block instead of removing those?

zengin:
@baywet ideally yes. Is the root cause still the same though? I don't have the local Java setup, I was hoping Java failures to go away as well because this was just a property added to V1, so I was relying on the Azure DevOps run.

I will have a quick look at the failure in the .trx because Azure DevOps is truncating error message.

zengin:
@baywet It looks like the property is added to the Java service library 21 days ago. Why does Raptor not picking up the changes?

https://github.com/microsoftgraph/msgraph-sdk-java/blame/d12ac2c598419566504032862cdf58ca1066f65a/src/main/java/com/microsoft/graph/models/extensions/Channel.java#L57

baywet:
@zengin after posting the message I spent a few minutes looking into that. I doesn't look like we're passing a specific version to the build process, which as far as I can remember means gradle will pull the latest version. Now it's possible it's not behaving like anticipated, running the test locally and running `gradle dependencies` in the locally generated project should tell you which version it's pulling.

zengin:
installing gradle on my WSL instance.

MIchaelMainer (approved):

azure-pipelines/java-v1-compile-tests-known-issues.yml
```diff
@@ -20,10 +20,12 @@ variables:
 steps:
 - template: java-templates/java-prepare.yml
 - template: common-templates/checkout.yml
+- template: java-templates/java-set-v1-version.yml
 - template: java-templates/java-replace-local-path.yml
   parameters:
     projectFileName: JavaV1KnownFailureTests
     javaPreviewLibPath: ''
+    javaLibVersion: $serviceLibraryVersion
```
baywet, line 28: Wouldn't this need parenthesis?

MIchaelMainer (approved):

nikithauc (approved):

zengin:
@baywet FYI, verified that the Java tests are passing as well after referencing the latest service library.

baywet:
thanks for taking the time to fix that and sorry I forgot about the default value :)

PR #133, opened 2020-12-10
Parallelize test runs

zengin:
- Set parallel scope to all for text expansion to run in parallel
- Set MaxCpuCount to 0 to leave cpu count decision to the runner. ([documentation](https://docs.microsoft.com/en-us/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file?view=vs-2019#runconfiguration-element))

### DIFF
--- a/msgraph-sdk-raptor-compiler-lib/MicrosoftGraphJavaCompiler.cs
+++ b/msgraph-sdk-raptor-compiler-lib/MicrosoftGraphJavaCompiler.cs
@@ -81,6 +81,14 @@ application {
 
         private static Versions? currentlyConfiguredVersion;
         private static readonly Lazy<int> currentExcutionFolder = new Lazy<int>(() => new Random().Next(0, int.MaxValue));
+        private static readonly object versionLock = new { };
+
+        private static void SetCurrentlyConfiguredVersion (Versions version)
+        {// we don't want to overwrite the build.gradle for each test, this prevents gradle from caching things and slows down build time
+            lock(versionLock) {
+                currentlyConfiguredVersion = version;
+            }
+        }
 
         public MicrosoftGraphJavaCompiler(string markdownFileName, string previewLibPath, string javaLibVersion, string javaCoreVersion)
         {
@@ -93,11 +101,13 @@ application {
         {
             var tempPath = Path.Combine(Path.GetTempPath(), "msgraph-sdk-raptor");
             Directory.CreateDirectory(tempPath);
-            var rootPath = Path.Combine(tempPath, "java" + currentExcutionFolder.Value, Guid.NewGuid().ToString());
+            var rootPath = Path.Combine(tempPath, "java" + currentExcutionFolder.Value);
             var sourceFileDirectory = Path.Combine(new string[] { rootPath }.Union(testFileSubDirectories).ToArray());
-
-            InitializeProjectStructure(version, rootPath).GetAwaiter().GetResult();
-
+            if (!currentlyConfiguredVersion.HasValue || currentlyConfiguredVersion.Value != version)
+            {
+                InitializeProjectStructure(version, rootPath).GetAwaiter().GetResult();
+                SetCurrentlyConfiguredVersion(version);
+            }
             File.WriteAllText(Path.Combine(sourceFileDirectory, "App.java"), codeSnippet); //could be async
             using var javacProcess = new Process
             {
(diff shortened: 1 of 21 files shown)


zengin:
Not a super improvement in the cloud. Some had no difference, some had almost 20% improvement. Best one was 43 minutes vs 53 minutes for Java Beta Tests.

I get much more improvement (almost 2x) in my local box, most likely due to the fact that I have more cores at my disposal there.

nikithauc (approved):

baywet:
@zengin Thanks for adding this improvement. I was curious about something: as the java tests write on disk, aren't we at a risk of parallel tests overwriting each other? I made sure the folder is created with some random name for each context, but my question is: is this parallel mode creating a different context for each parallel instance?

zengin:
@baywet, it looks like it is writing to the same folder for each test. I will fix that. I assumed that it was safe for parallelism when MaxCount:4 was added.

baywet:
As far as I can remember the maxcount option sets a new context per process (per CPU core). Which I believe means a separate folder per process as it's implemented today.

zengin:
@baywet I lose the local caching of build.gradle file with the latest change, but parallel run seems to be still paying off. I still get 2x improvement in Java local run on MacOS. I will see how it looks like in the cloud with less cores.

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphJavaCompiler.cs
```diff
@@ -126,6 +116,13 @@ public CompilationResultsModel CompileSnippet(string codeSnippet, Versions versi
                 javacProcess.Kill(true);
             var stdOutput = javacProcess.StandardOutput.ReadToEnd(); //could be async
             var stdErr = javacProcess.StandardError.ReadToEnd(); //could be async
+
+            var config = AppSettings.Config();
+            if (bool.Parse(config.GetSection("DeleteJavaTempDirectory").Value))
+            {
+                Directory.Delete(rootPath, true);
```
baywet, line 123: doing this on each test will pull the dependencies on each test which will significantly increase runtime duration

zengin: Well, it didn't. So I assume there is some local cache of the dependencies. Would you mind double checking the perf on  Windows when your machine has some idle time? I don't have the Java setup on my Windows machine.

zengin: Never mind, I own this change, I will use this opportunity to setup Java on Windows. I will not merge it if it causes a slow down on Windows.

msgraph-sdk-raptor-compiler-lib/appsettings.json
```diff
@@ -24,5 +24,6 @@
   },
   "IsLocalRun": false,
   "GenerateLinqPadOutputInLocalRun": true,
-  "LocalRootGitDirectory": "C:/github"
-}
\ No newline at end of file
+  "LocalRootGitDirectory": "C:/github",
```
MIchaelMainer, line 27: Under which circumstances is git at `c:/github`?

`where.exe git.exe` can provide the git.exe location in bash and PowerShell.

MIchaelMainer (approved):

msgraph-sdk-raptor-compiler-lib/appsettings.json
zengin: This is the root directory where microsoft-graph-docs is checked out. I should do a better job in naming this parameter.

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphJavaCompiler.cs
zengin: My Windows machine went crazy with the parallel read/writes. Even though Linux/Mac performs really well with the new code, I am reverting the changes for Java for the sake of platform support.

MIchaelMainer (approved):

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphJavaCompiler.cs
zengin: I will leave the MacOS numbers here for further reference in case we revisit this implementation:
```
dotnet test JavaV1Tests  772.93s user 101.49s system 41% cpu 35:28.70 total
dotnet test JavaV1Tests -v:n  1107.22s user 241.05s system 154% cpu 14:34.25 total

dotnet test JavaBetaTests -v:n  1257.68s user 168.12s system 54% cpu 43:13.48 total
dotnet test JavaBetaTests -v:n  1874.67s user 400.32s system 160% cpu 23:33.38 total
```

baywet:
@zengin It's roughly the same apparently on the cloud build. I'm afraid it'll slow down local runs considerably though from what I can remember when I originally worked on that.

zengin:
@baywet, yeah. I reverted the changes in Java. For C#, it makes a difference in the local as well. So I will keep those.

baywet (approved):

PR #138, opened 2020-12-21
- fixes an issue where the javac process could potentially deadlock with raptor

baywet:
more information https://stackoverflow.com/questions/139593/processstartinfo-hanging-on-waitforexit-why
At the time I thought this was just gradle going in some kind of infinite loop, but it looks like it could be a simple deadlock.


zengin (approved):
I wonder if this was the reason why it went crazy when I tried the parallelization changes on Windows. I am almost tempted to retry :)

baywet:
@zengin It surely had a negative impact. The other thing is the dependency on a folder, so parallel workers must work on separate folders.

Issue #164, opened 2021-02-02
Determine when to use delegated vs application permissions


zengin:
We need to figure out which type of permissions are needed for each snippet.

Documentation has this information in markdown format. 
[AB#7969](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7969)

zengin:
Fixed in #416

PR #166, opened 2021-02-03
Infrastructure for fetching delegated permissions

### LINKED ISSUE #165: Create infrastructure to fetch delegated permissions
1. Extract generated URL from the snippet
2. Call DevX API to get snippets
3. Extend appsettings to hold username/password 
[AB#7971](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7971)

zengin:
- Adds a new method into the snippet decoration to extract HttpRequestMessage and URL.
- Modifies snippet to return `HttpRequestMessage` using the template above.
- Has a simple logic deciding on delegated vs. application permissions. Full extent of this logic will be covered by #164 
- Calls DevX API to fetch required permissions after getting the URL from compiled snippet.
- Extends `appsettings.json` to hold authority, username and password. These are required to fetch a token with delegated permissions.
- Fixes #165

cc: @irvinesunday (please review if there are any issues with calling the DevX API)

### DIFF
--- a/TestsCommon/CSharpTestRunner.cs
+++ b/TestsCommon/CSharpTestRunner.cs
@@ -118,7 +118,7 @@ public class GraphSDKTest
 
             // Compile Code
             var microsoftGraphCSharpCompiler = new MicrosoftGraphCSharpCompiler(testData.FileName, testData.DllPath);
-            var executionResultsModel = await microsoftGraphCSharpCompiler.ExecuteSnippet(codeToCompile, testData.Version);
+            var executionResultsModel = await microsoftGraphCSharpCompiler.ExecuteSnippet(codeToCompile, testData.Version).ConfigureAwait(false);
             var compilationOutputMessage = new CompilationOutputMessage(
                 executionResultsModel.CompilationResult,
                 codeToCompile,
--- a/TestsCommon/CSharpTestRunner.cs
+++ b/TestsCommon/CSharpTestRunner.cs
@@ -40,6 +40,11 @@ public class GraphSDKTest
         //insert-code-here
         #endregion
     }
+
+    public HttpRequestMessage GetRequestMessage(IAuthenticationProvider authProvider)
+    {
+        //return-request-message
+    }
 }";
 
         /// <summary>
@@ -177,6 +182,32 @@ public class GraphSDKTest
             return codeToCompile;
         }
 
+        /// <summary>
+        /// Modifies snippet to return HttpRequestMessage object so that we can extract the generated URL
+        /// </summary>
+        /// <param name="codeSnippet">code snippet</param>
+        /// <returns>Code snippet that returns an HttpRequstMessage</returns>
+        private static string ReturnHttpRequestMessage(string codeSnippet)
+        {
+            string resultVariable = null;
+            try
+            {
+                resultVariable = ResultVariableRegex.Match(codeSnippet).Groups[1].Value;
+            }
+            catch (Exception e)
+            {
+                Assert.Fail("result variable is not found!" + Environment.NewLine + e.Message);
+            }
+
+            codeSnippet = codeSnippet.Replace("await graphClient", "graphClient")
+                .Replace(".GetAsync();", $@".GetHttpRequestMessage();
+
+        return {resultVariable};");
+
+            return codeSnippet;
+        }
+
+
         private static (string, string) GetCodeToCompile(string fileContent, Func<string, string> postTransform = null)
         {
             var match = CSharpSnippetRegex.Match(fileContent);
@@ -189,6 +220,10 @@ public class GraphSDKTest
                 .Replace("\r\n\r\n\r\n", "\r\n\r\n");       // do not have two consecutive empty lines
 
             var codeToCompile = BaseTestRunner.ConcatBaseTemplateWithSnippet(codeSnippetFormatted, SDKShellTemplate);
+
+            // have another tranformation to insert GetRequestMessage method
+            codeToCompile = codeToCompile.Replace("//return-request-message", "//insert-code-here");
+            codeToCompile = BaseTestRunner.ConcatBaseTemplateWithSnippet(ReturnHttpRequestMessage(codeSnippetFormatted), codeToCompile);
 
             if (postTransform == null)
             {
--- a/TestsCommon/CSharpTestRunner.cs
+++ b/TestsCommon/CSharpTestRunner.cs
@@ -186,7 +186,8 @@ public class GraphSDKTest
         /// Modifies snippet to return HttpRequestMessage object so that we can extract the generated URL
         /// </summary>
         /// <param name="codeSnippet">code snippet</param>
-        /// <returns>Code snippet that returns an HttpRequstMessage</returns>
+        /// <returns>Code snippet that returns an HttpRequestMessage</returns>
+
         private static string ReturnHttpRequestMessage(string codeSnippet)
         {
             string resultVariable = null;
--- a/msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
+++ b/msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
@@ -19,8 +19,8 @@ using System.Threading.Tasks;
 using System.Reflection;
 using System.Text.RegularExpressions;
 using System.Security;
-using System.Net;
 using System.Text.Json;
+using Microsoft.Extensions.Configuration;
 
 namespace MsGraphSDKSnippetsCompiler
 {
@@ -129,7 +129,7 @@ namespace MsGraphSDKSnippetsCompiler
                 {
                     var requiresDelegatedPermissions = RequiresDelegatedPermissions(codeSnippet);
                     var config = AppSettings.Config();
-                    var clientId = config.GetSection("ClientID").Value;
+                    var clientId = GetNonEmptyValue(config, "ClientID");
 
                     dynamic instance = assembly.CreateInstance("GraphSDKTest");
                     IAuthenticationProvider authProvider;
@@ -137,12 +137,12 @@ namespace MsGraphSDKSnippetsCompiler
                     if (requiresDelegatedPermissions)
                     {
                         // delegated permissions
-                        HttpRequestMessage httpRequestMessage = instance.GetRequestMessage(null);
-                        var scopes = GetScopes(httpRequestMessage);
-                        var authority = config.GetSection("Authority").Value;
-                        var username = config.GetSection("Username").Value;
-                        var password = config.GetSection("Password").Value;
-                        var token = GetATokenForGraph(clientId, authority, username, password, scopes).ConfigureAwait(false).GetAwaiter().GetResult();
+                        using var httpRequestMessage = instance.GetRequestMessage(null);
+                        var scopes = await GetScopes(httpRequestMessage);
+                        var authority = GetNonEmptyValue(config, "Authority");
+                        var username = GetNonEmptyValue(config, "Username");
+                        var password = GetNonEmptyValue(config, "Password");
+                        var token = await GetATokenForGraph(clientId, authority, username, password, scopes).ConfigureAwait(false);
                         authProvider = new DelegateAuthenticationProvider(async request =>
                         {
                             request.Headers.Authorization = new System.Net.Http.Headers.AuthenticationHeaderValue("Bearer", token);
@@ -151,8 +151,8 @@ namespace MsGraphSDKSnippetsCompiler
                     else
                     {
                         // application permissions
-                        var tenantId = config.GetSection("TenantID").Value;
-                        var clientSecret = config.GetSection("ClientSecret").Value;
+                        var tenantId = GetNonEmptyValue(config, "TenantID");
+                        var clientSecret = GetNonEmptyValue(config, "ClientSecret");
                         IConfidentialClientApplication confidentialClientApp = ConfidentialClientApplicationBuilder
                             .Create(clientId)
                             .WithTenantId(tenantId)
@@ -166,7 +166,8 @@ namespace MsGraphSDKSnippetsCompiler
                 }
                 catch (Exception e)
                 {
-                    exceptionMessage = e.Message + Environment.NewLine + e.InnerException.Message;
+                    var innerExceptionMessage = e.InnerException?.Message ?? string.Empty;
+                    exceptionMessage = e.Message + Environment.NewLine + innerExceptionMessage;
                     if (!bool.Parse(AppSettings.Config().GetSection("IsLocalRun").Value))
                     {
                         exceptionMessage = AuthHeaderRegex.Replace(exceptionMessage, AuthHeaderReplacement);
@@ -182,7 +183,7 @@ namespace MsGraphSDKSnippetsCompiler
         /// </summary>
         /// <param name="httpRequestMessage"></param>
         /// <returns></returns>
-        static string[] GetScopes(HttpRequestMessage httpRequestMessage)
+        static async Task<string[]> GetScopes(HttpRequestMessage httpRequestMessage)
         {
             var path = httpRequestMessage.RequestUri.LocalPath;
             var versionSegmentLength = "/v1.0".Length;
@@ -191,13 +192,17 @@ namespace MsGraphSDKSnippetsCompiler
                 path = path[versionSegmentLength..];
             }
 
-            var webClient = new WebClient();
-            webClient.Headers.Add("Accept-Language", "en-US");
-            var url = $"https://graphexplorerapi.azurewebsites.net/permissions?requesturl={path}&method={httpRequestMessage.Method}";
-            var responseString = webClient.DownloadString(url);
-            var json = JsonSerializer.Deserialize<Scope[]>(responseString);
+            using var httpClient = new HttpClient();
 
-            return json.ToList().Select(x => "https://graph.microsoft.com/" + x.value).ToArray();
+            using var scopesRequest = new HttpRequestMessage(HttpMethod.Get, $"https://graphexplorerapi.azurewebsites.net/permissions?requesturl={path}&method={httpRequestMessage.Method}");
+            scopesRequest.Headers.Add("Accept-Language", "en-US");
+
+            using var response = await httpClient.SendAsync(scopesRequest).ConfigureAwait(false);
+            var responseString = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+
+            var scopes = JsonSerializer.Deserialize<Scope[]>(responseString);
+
+            return scopes.ToList().Select(x => $"https://graph.microsoft.com/{ x.value }").ToArray();
         }
 
         /// <summary>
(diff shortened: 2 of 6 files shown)


TestsCommon/CSharpTestRunner.cs
```diff
@@ -40,6 +40,11 @@ public async Task Main(IAuthenticationProvider authProvider)
         //insert-code-here
         #endregion
     }
+
+    public HttpRequestMessage GetRequestMessage(IAuthenticationProvider authProvider)
```
nikithauc, line 44: Can you please add a summary to this method?

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -155,6 +177,59 @@ public async Task<ExecutionResultsModel> ExecuteSnippet(string codeSnippet, Vers
...
+        /// <param name="username">username of the user for which the token is requested</param>
+        /// <param name="password">password of the user for which the token is requested</param>
+        /// <param name="scopes">requested scopes in the token</param>
+        /// <returns>token for the given context</returns>
+        static async Task<string> GetATokenForGraph(string clientId, string authority, string username, string password, string[] scopes)
+        {
+            var app = PublicClientApplicationBuilder.Create(clientId).WithAuthority(authority).Build();
+
+            var securePassword = new SecureString();
+            password.ToList().ForEach(c => securePassword.AppendChar(c));
```
nikithauc, line 217: What does this do?

msgraph-sdk-raptor-compiler-lib/appsettings.json
```diff
@@ -4,5 +4,8 @@
   "DocsRepoCheckoutDirectory": "C:/github",
```
ddyett, line 4: not this PR, but you'll probably want to clean up windows specific paths.

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -155,6 +177,59 @@ public async Task<ExecutionResultsModel> ExecuteSnippet(string codeSnippet, Vers
...
+
+            try
+            {
+                var result = await app.AcquireTokenByUsernamePassword(scopes, username, securePassword).ExecuteAsync();
+                return result.AccessToken;
+            }
+            catch (Exception e)
+            {
+                var PrefixLength = "https://graph.microsoft.com/".Length;
```
ddyett, line 226: NIT: your naming convention is slightly different here.  Should be prefixLength

TestsCommon/CSharpTestRunner.cs
```diff
@@ -177,6 +182,32 @@ private static string CaptureUriAndHeadersInException(string codeToCompile)
             return codeToCompile;
         }
 
+        /// <summary>
+        /// Modifies snippet to return HttpRequestMessage object so that we can extract the generated URL
+        /// </summary>
+        /// <param name="codeSnippet">code snippet</param>
+        /// <returns>Code snippet that returns an HttpRequstMessage</returns>
```
ddyett, line 189:
```suggestion
        /// <returns>Code snippet that returns an HttpRequestMessage</returns>

```

TestsCommon/CSharpTestRunner.cs
zengin: It is intentional not to have a comment here, because ADO limits the number of lines shown in test results. If the compilation fails, I print the snippet that we are compiling and running. I don't want to truncate because of a comment.

The purpose of this method is given in the function that populates it.

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
zengin:
```suggestion

            // convert plain password into a secure string.
            password.ToList().ForEach(c => securePassword.AppendChar(c));
```

msgraph-sdk-raptor-compiler-lib/appsettings.json
zengin: it was just a convenience for myself and we were replacing this in linux runs. But I might as well remove it.
```suggestion
  "DocsRepoCheckoutDirectory": "",
```

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -155,6 +177,59 @@ public async Task<ExecutionResultsModel> ExecuteSnippet(string codeSnippet, Vers
...
+                var scopeShortNames = scopes.Select(s => s[PrefixLength..]).ToArray();
```
zengin, line 227:
```suggestion
                var prefixLength = "https://graph.microsoft.com/".Length;
                var scopeShortNames = scopes.Select(s => s[prefixLength..]).ToArray();
```

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -124,20 +127,39 @@ public async Task<ExecutionResultsModel> ExecuteSnippet(string codeSnippet, Vers
...
-                        .WithTenantId(tenantId)
-                        .WithClientSecret(clientSecret)
-                        .Build();
-                    authProvider = new ClientCredentialProvider(confidentialClientApp, DefaultAuthScope);
+                    if (requiresDelegatedPermissions)
+                    {
+                        // delegated permissions
+                        HttpRequestMessage httpRequestMessage = instance.GetRequestMessage(null);
+                        var scopes = GetScopes(httpRequestMessage);
+                        var authority = config.GetSection("Authority").Value;
```
baywet, line 142: null propagation?

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -124,20 +127,39 @@ public async Task<ExecutionResultsModel> ExecuteSnippet(string codeSnippet, Vers
...
-                    IConfidentialClientApplication confidentialClientApp = ConfidentialClientApplicationBuilder
-                        .Create(clientId)
-                        .WithTenantId(tenantId)
-                        .WithClientSecret(clientSecret)
-                        .Build();
-                    authProvider = new ClientCredentialProvider(confidentialClientApp, DefaultAuthScope);
+                    if (requiresDelegatedPermissions)
+                    {
+                        // delegated permissions
+                        HttpRequestMessage httpRequestMessage = instance.GetRequestMessage(null);
```
baywet, line 140:
```suggestion
                        using var httpRequestMessage = instance.GetRequestMessage(null);
```

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -124,20 +127,39 @@ public async Task<ExecutionResultsModel> ExecuteSnippet(string codeSnippet, Vers
...
+                        var scopes = GetScopes(httpRequestMessage);
+                        var authority = config.GetSection("Authority").Value;
+                        var username = config.GetSection("Username").Value;
+                        var password = config.GetSection("Password").Value;
+                        var token = GetATokenForGraph(clientId, authority, username, password, scopes).ConfigureAwait(false).GetAwaiter().GetResult();
```
baywet, line 145:
```suggestion
                        var token = await GetATokenForGraph(clientId, authority, username, password, scopes).ConfigureAwait(false);
```

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -124,20 +127,39 @@ public async Task<ExecutionResultsModel> ExecuteSnippet(string codeSnippet, Vers
...
+                        authProvider = new DelegateAuthenticationProvider(async request =>
+                        {
+                            request.Headers.Authorization = new System.Net.Http.Headers.AuthenticationHeaderValue("Bearer", token);
+                        });
+                    }
+                    else
+                    {
+                        // application permissions
+                        var tenantId = config.GetSection("TenantID").Value;
```
baywet, line 154: null prop?

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -155,6 +177,61 @@ public async Task<ExecutionResultsModel> ExecuteSnippet(string codeSnippet, Vers
...
+        static string[] GetScopes(HttpRequestMessage httpRequestMessage)
+        {
+            var path = httpRequestMessage.RequestUri.LocalPath;
+            var versionSegmentLength = "/v1.0".Length;
+            if (path.StartsWith("/v1.0") || path.StartsWith("/beta"))
+            {
+                path = path[versionSegmentLength..];
+            }
+
+            var webClient = new WebClient();
```
baywet, line 194:
```suggestion
            using var webClient = new WebClient();
```

baywet, line 194: out of curiosity, is there a reason you are using webclient and not http client?

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -155,6 +177,61 @@ public async Task<ExecutionResultsModel> ExecuteSnippet(string codeSnippet, Vers
...
+            webClient.Headers.Add("Accept-Language", "en-US");
+            var url = $"https://graphexplorerapi.azurewebsites.net/permissions?requesturl={path}&method={httpRequestMessage.Method}";
+            var responseString = webClient.DownloadString(url);
+            var json = JsonSerializer.Deserialize<Scope[]>(responseString);
```
baywet, line 198: I don't believe this variable stores json, rename?

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -155,6 +177,61 @@ public async Task<ExecutionResultsModel> ExecuteSnippet(string codeSnippet, Vers
...
+
+            return json.ToList().Select(x => "https://graph.microsoft.com/" + x.value).ToArray();
```
baywet, line 200:
```suggestion
            return json.ToList().Select(x => $"https://graph.microsoft.com/{x.value}").ToArray();
```

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -155,6 +177,61 @@ public async Task<ExecutionResultsModel> ExecuteSnippet(string codeSnippet, Vers
...
+        /// <param name="authority">token authority, such as: https://login.microsoftonline.com/contoso.onmicrosoft.com</param>
+        /// <param name="username">username of the user for which the token is requested</param>
+        /// <param name="password">password of the user for which the token is requested</param>
+        /// <param name="scopes">requested scopes in the token</param>
+        /// <returns>token for the given context</returns>
+        static async Task<string> GetATokenForGraph(string clientId, string authority, string username, string password, string[] scopes)
+        {
+            var app = PublicClientApplicationBuilder.Create(clientId).WithAuthority(authority).Build();
+
+            var securePassword = new SecureString();
```
baywet, line 216:
```suggestion
            using var securePassword = new SecureString();
```

msgraph-sdk-raptor-compiler-lib/Models/Scope.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace MsGraphSDKSnippetsCompiler.Models
+{
+    /// <summary>
+    /// Model from https://github.com/microsoftgraph/microsoft-graph-explorer-api/blob/dev/GraphExplorerPermissionsService/Models/ScopeInformation.cs
+    /// </summary>
+    public class Scope
```
baywet, line 6: use record instead?

scripts/transformAppSettings.ps1
```diff
@@ -4,7 +4,10 @@ param(
     [string]$DocsRepoCheckoutDirectory = "C:/github",
     [Parameter(Mandatory=$true)][string]$TenantID,
     [Parameter(Mandatory=$true)][string]$ClientID,
-    [Parameter(Mandatory=$true)][string]$ClientSecret
+    [Parameter(Mandatory=$true)][string]$ClientSecret,
```
baywet, line 7: isn't making those required going to block the java run until we update java pipelines? or are you planning to update the java pipelines as well?

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -155,6 +177,61 @@ public async Task<ExecutionResultsModel> ExecuteSnippet(string codeSnippet, Vers
...
+            try
+            {
+                var result = await app.AcquireTokenByUsernamePassword(scopes, username, securePassword).ExecuteAsync();
+                return result.AccessToken;
+            }
+            catch (Exception e)
+            {
+                var prefixLength = "https://graph.microsoft.com/".Length;
+                var scopeShortNames = scopes.Select(s => s[prefixLength..]).ToArray();
+                throw new AggregateException(new AggregateException("scopes: " + string.Join(", ", scopeShortNames), e));
```
baywet, line 230: Why are you double wrapping the exception? maybe use Task.FromException instead?

baywet (changes requested):

scripts/transformAppSettings.ps1
zengin: This script is not used in the Java pipelines. You can still have appsettings with these values empty/missing.

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
zengin: `GetSection()` never returns `null`: https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationroot.getsection?view=dotnet-plat-ext-5.0#remarks

But I will add a check if they are empty.

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
zengin: Less code for simple requests. Do I miss much by not using http client?

msgraph-sdk-raptor-compiler-lib/Models/Scope.cs
zengin: I have tried that and it didn't play well with the Json Deserializer. I will give it another try because I don't remember the error ;)

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
zengin: It was to match with the exceptions I was getting elsewhere so that the onion peeling is the same for exceptions thrown at different places, but that had changed to be single level of aggregate after async changes.

I will remove one layer from here.

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
zengin: ah, leftover from the old implementation, thanks!

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
zengin: Interesting, this was taken from the documentation. I guess I should let them know as well :)

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
baywet: you are using a blocking API instead of an async one for the downloadString :)
Plus I believe WebClient on dotnet core is essentially a wrapper around HttpClient and it's not going to evolve much going forward

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
baywet: good to know. There's probably a bunch of application still in production from my former jobs that I wrote with null propagation operators like I'm suggesting :D

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
baywet: I believe the reason why you need so much wrapping is because you were not using async/await at a couple of places and the AggregateException were not being debounced for you.

msgraph-sdk-raptor-compiler-lib/Models/Scope.cs
zengin: It worked now :D

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
zengin: I am now checking for null on innerException, so no more wrapping.

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -277,5 +282,22 @@ private static bool RequiresDelegatedPermissions(string codeSnippet)
...
+
+        /// <summary>
+        /// Extracts the configuration value, throws if empty string
+        /// </summary>
+        /// <param name="config">configuration</param>
+        /// <param name="key">lookup key</param>
+        /// <returns>non-empty configuration value if found</returns>
+        private static string GetNonEmptyValue(IConfigurationRoot config, string key)
+        {
+            var value = config.GetSection(key).Value;
```
baywet, line 294: return config?.GetSection(key)?.Value ?? throw new Exception($"Value for {key} is not found in appsettings.json");

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
baywet: yeah I'm not sure if it's actually doing anything. But just in case it's storing somewhere with the help of a TPM in some cases...

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
zengin: I am not doing null check intentionally on config since this is a private method and as mentioned earlier GetSection never returns null.

zengin: And only checking `null` narrows the functionality. Empty string is the default value in the JSON, so checking on that informs the test runner who hasn't configured these settings.

baywet: right, I got mixed up with typescript where the !someString will also check for empty. Sorry about that.

baywet (approved):

PR #187, opened 2021-02-20
Add a permission update script for AAD application manifest 

### LINKED ISSUE #186: Create a script for modifying application manifest file
Application manifest files can be downloaded from/uploaded to the Azure Portal.

We know the permission list for `Microsoft.Graph`.

Create a script that
- downloads permissions list from https://raw.githubusercontent.com/microsoftgraph/microsoft-graph-devx-content/dev/permissions/permissions-descriptions.json
- Selects read-only permissions from that list
- Applies those permissions to the manifest file in the expected format.

This will allow us to easily create new applications with desired permissions.

Admin consent still requires a click in the portal but it is once per all the permissions. With this script, we won't need to check the boxes in the UI for each permission in the list. 
[AB#8230](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8230)

zengin:
Fixes #186

- Modifies an AAD application manifest to have all read-only permissions to `Microsoft.Graph`:
  - Downloads permissions list from https://raw.githubusercontent.com/microsoftgraph/microsoft-graph-devx-content/dev/permissions/permissions-descriptions.json
  - Selects read-only permissions from that list
  - Applies those permissions to the manifest file in the expected format.

This allows us to easily create new applications with desired permissions.

Admin consent still requires a click in the portal but it is once per all the permissions. With this script, we won't need to check the boxes in the UI for each permission in the list.

cc: @Jackson-Woods @darrelmiller @pixieofhugs

### DIFF
--- a/scripts/applyReadPermissionsToManifest.ps1
+++ b/scripts/applyReadPermissionsToManifest.ps1
@@ -1,0 +1,108 @@
+# Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+
+<#
+.SYNOPSIS
+  This script updates an AAD application manifest file to have all read-only permissions to Microsoft.Graph
+.DESCRIPTION
+  - Fetches all Microsoft.Graph permissions
+  - Selects read-only permissions
+  - Updates application manifest file to have read-only permissions
+.EXAMPLE
+  In-place manifest update
+  PS C:\> applyReadOnlyPermissionsToManifest.ps1 -manifestPath <path_to_manifest>
+
+  You can also specify optional outputPath if you don't want in-place update
+  PS C:\> applyReadOnlyPermissionsToManifest.ps1 -manifestPath <path_to_manifest> -outputPath <path_to_output>
+.NOTES
+  Manifest file can be downloaded from Azure Portal -> Application -> Manifest
+  On the same Manifest page, you can upload the manifest file modified with this script
+  When you switch back to API Permissions page, you will see that all read-only
+  Microsoft.Graph permissions will be given to your application
+  For admin consent, you will still need to click on "Grant admin consent for Contoso" button.
+#>
+
+Param(
+  [Parameter(Mandatory)]
+  [string]$manifestPath,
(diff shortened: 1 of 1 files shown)


scripts/applyReadPermissionsToManifest.ps1
```diff
@@ -0,0 +1,107 @@
...
+  ConvertFrom-Json
+
+if (!$manifest.requiredResourceAccess.resourceAppId)
+{
+  Write-Error "Given file is not a manifest file that we can modify. It needs to have requiredResourcesAccess.resourceAppId field already defined!"
+  return;
+}
+
+# fetch permissions from microsoft-graph-devx-content
+$permissionsFile = "https://raw.githubusercontent.com/microsoftgraph/microsoft-graph-devx-content/dev/permissions/permissions-descriptions.json"
```
baywet, line 45: could that be the default value of a parameter instead? In case we'd ever want to use another source for testing purposes.

scripts/applyReadPermissionsToManifest.ps1
```diff
@@ -0,0 +1,107 @@
...
+$permissions = (Invoke-WebRequest -Uri $permissionsFile -UseBasicParsing).Content | ConvertFrom-Json
+
+# filter read only permissions
+$readOnlyDelegatedPermissions = $permissions.delegatedScopesList |
+  Where-Object { $_.value.Contains("Read") -and !$_.value.Contains("Write") }
```
baywet, line 50: Constants?

baywet (approved):

scripts/applyReadPermissionsToManifest.ps1
zengin: Thanks, made this a param.

scripts/applyReadPermissionsToManifest.ps1
zengin: moved repeated code into an inline function, so these constants appear only once in the code now. Thanks!

nikithauc (approved):

Issue #196, opened 2021-02-26
Add/Remove known issues


zengin:
This issue is to track changes we are making into the known issues list.

PR #197, opened 2021-02-26
Add/remove known issues

zengin:
- Removed passing tests from the known issues list.
- Added more analysis. Decided to mark consistently failing tests as "needs analysis" if I don't know the root cause yet.
- @baywet I used the same "needs analysis" tag for cases where
  - there was a common blocker for both Java and C#. i.e. it was listed under common issues.
  - the common blocker (either docs or metadata) is fixed, but Java test continues to fail, implying that there is another root cause on top of the initial blocker.

Part of #196

### DIFF
--- a/TestsCommon/TestDataGenerator.cs
+++ b/TestsCommon/TestDataGenerator.cs
@@ -55,6 +55,8 @@ namespace TestsCommon
             " See https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/485";
         private const string EnumsAreNotHandled = "Handling of enums is not correct" +
             " See https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/482";
+        private const string SameBlockNames = "Same block names indeterministic snippet generation" +
+            " See https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/463";
         #endregion
 
         #region Needs analysis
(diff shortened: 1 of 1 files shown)


TestsCommon/TestDataGenerator.cs
```diff
@@ -39,19 +40,27 @@ public static class KnownIssues
...
         #region Snipppet Generation Issues
-        private const string SnippetGenerationFlattens = "Snippet generation flattens the nested Odata queries." +
-            " See https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/287";
         private const string SnippetGenerationAdditionalData = "Open types should use additional data for non-existent properties." +
             " See https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/296";
         private const string SnippetGenerationCreateAsyncSupport = "Snippet generation doesn't use CreateAsync" +
             " See https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/301";
         private const string SnippetGenerationRequestObjectDisambiguation = "Snippet generation should rename objects that end with Request to end with RequestObject" +
             " See https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/298";
+        private const string RegularPropertiesAreNotHandled = "We should build a custom request with URL as this is not supported in SDK. " +
```
MIchaelMainer, line 54: prepend: We don't generate request builders for URL navigation to structural properties.

MIchaelMainer (approved):

MIchaelMainer (approved):

baywet:
@zengin is the Java test suite failing because we're reaching one hour of run time or is it because of something different?

zengin:
@baywet It is running out of time especially on beta tests as the number of tests increased significantly and 1 hour doesn't work anymore. It would be nice if we can go back to perf for Java.

For this particular PR, latest commit didn't have a code change, so I relied on previous run while merging.

PR #199, opened 2021-02-26
Staging AGS pipelines

### LINKED ISSUE #198: Add staging pipelines for running tests that are not known issues
This is for running the tests upon deployment to staging by AGS.

- Have a separate staging end-to-end pipeline, but only run the tests that are not known issues so that the baseline is clear for AGS folks.
- Keep existing daily staging runs to follow the progress on known issues. 
[AB#8326](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8326)

zengin:
Added two new pipelines to be triggered by AGS upon deployment. These pipelines exclude known issues from the test suite. They are expected to pass with 100% for each deployment.

I still keep the original staging pipelines that run daily to keep track of known issues. I am also not running snippet generation step for these pipelines to avoid randomization from docs repo check-ins.

Fixes #198

Also added one more known issue (part of #196)


baywet (approved):
Out of curiosity. Couldn't we have those pipelines on the AGS repo so they become a PR validation gate? They could checkout raptor, reuse all the stuff we have here in terms of pipelines etc.

zengin:
@baywet, that will be the next step. There are a few blockers that we are working on for enabling that.

PR #203, opened 2021-03-03
Fix known issue messages

zengin:
Some of the known issues are fixed, but there were other reasons for them to continue failing to compile. Fixed those error messages.

Sorted C# list by issue type.

Part of #196

### DIFF
--- a/TestsCommon/TestDataGenerator.cs
+++ b/TestsCommon/TestDataGenerator.cs
@@ -25,6 +25,8 @@ namespace TestsCommon
         private const string PutAsyncIsNotSupported = "PutAsync methods are not auto generated.\r\n"
             + "https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/844";
         private const string StreamRequestDoesNotSupportDelete = "Stream requests only support PUT and GET.";
+        private const string DeleteAsyncIsNotSupportedForReferences = "DeleteAsync is not supported for reference collections\r\n"
+            + "https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/issues/471";
 
         #endregion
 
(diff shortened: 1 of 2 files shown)


TestsCommon/TestDataGenerator.cs
```diff
@@ -199,61 +200,77 @@ private static string GetMethodWrongMessage(string docsMethod, string expectedMe
...
-                { "put-privilegedrolesettings-csharp-Beta-compiles", new KnownIssue(SnippetGeneration, SnippetGenerationCreateAsyncSupport) },
-                { "put-regionalandlanguagesettings-csharp-Beta-compiles", new KnownIssue(SnippetGeneration, SnippetGenerationCreateAsyncSupport) },
+
+                {$"get-rooms-in-roomlist-csharp-{version}-compiles", new KnownIssue(SDK, "SDK doesn't generate type segment in OData URL. https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_workitems/edit/4997") },
+
+                { "put-privilegedrolesettings-csharp-Beta-compiles", new KnownIssue(SDK, PutAsyncIsNotSupported) },
+                { "put-regionalandlanguagesettings-csharp-Beta-compiles", new KnownIssue(SDK, PutAsyncIsNotSupported) },
+                { "team-put-schedule-csharp-Beta-compiles", new KnownIssue(SDK, PutAsyncIsNotSupported) },
+                { "update-organizationalbrandingproperties-csharp-Beta-compiles", new KnownIssue(SDK, PutAsyncIsNotSupported) },
+                {$"timeoff-put-csharp-{version}-compiles", new KnownIssue(SDK, PutAsyncIsNotSupported) },
```
MIchaelMainer, line 240: Why is the version is injected in some of these and not in other ones?

MIchaelMainer (approved):

TestsCommon/TestDataGenerator.cs
zengin: It means that it is a known issue in both V1 and Beta, so if you ask for `timeoff-put-csharp-V1-compiles` or `timeoff-put-csharp-Beta-compiles`, both will be returned as known issue. The others hard-code the version because they are only in Beta or only in V1.

PR #212, opened 2021-03-16
main update

baywet:
- java v3
- deps updates
- other minor changes


zengin (approved):

baywet:
@zengin I can see that this branch policy bumps things to 2 approvers instead of one. No issues with it, I just want to make sure this was intended?

zengin:
@baywet thanks for bringing that to attention. It was not intended.

baywet:
interesting, because the policy was changed after the vote, the auto-merge was stuck...

PR #219, opened 2021-03-30
Initial TenantSetup Infrastructure Using Pester

georgend:
- For each Entity such as Applications or Users, there is a folder for each HTTP request.
- The tests reads the `{Entity}/GET/{entity}.http` request to get the URL to list all entities. 
- If the entity in `{Entity}/POST/{entity}.http` does not exist, its create and stored in "RaptorIdentifiers.json", otherwise its identifier is stored in the Tree structure inside RaptorIdentifiers.json. 
- The tests use [ApiDoctor's HttpParser](https://github.com/microsoftgraph/apidoctor/blob/master/ApiDoctor.Validation/Http/HttpParser.cs) to convert the http data into an easy to use data model. Will download ApiDoctor from github into a local folder. 
- Re-Running the tests with same HttpData results in updates to RaptorIdentifiers.json. 

**How to Run**
`Install-Module -Name Pester -Force -SkipPublisherCheck`
`Invoke-Pester .\PopulateTenantFromHttpData.ps1 -Output Detailed`

**Requirements**
Requires an `env.json` file in the root of TenantSetup folder. `env.template.json` contains sample structure. 
***Certificate*** should be in LocalMachine store, and uploaded to testing tenant under the registered application.


scripts/TenantSetup/Populate-Tenant.ps1
```diff
@@ -0,0 +1,244 @@
...
+    $loadEnvironment = Join-Path $PSScriptRoot 'loadEnv.ps1'
+    $installTools = Join-Path $PSScriptRoot 'Install-Tools.ps1'
+    . $loadEnvironment
+    . $installTools
+
+    Install-MicrosoftGraph
+    Install-Az
+
+    Connect-MgGraph
+    #Connect-AzAccount
```
MIchaelMainer, line 25: Is there a reason to leave this commented out? Either remove, or provide a comment why this is here.

scripts/TenantSetup/Populate-Tenant.ps1
```diff
@@ -0,0 +1,244 @@
...
+}
+
+<#
+    Contacts
+#>
+Describe "Contacts" {
+    It "Create Contact Folder If Not Exists" {
+        #Contacts
+        #Create If Not Exists a Contact Folder
+        $contactFolder = Get-MgUserContactFolder
```
MIchaelMainer, line 66: I'm curious, does Get-MgUserContactFolder use the context of the user who ran Connect-MgGraph?

scripts/TenantSetup/loadEnv.ps1
```diff
@@ -0,0 +1,40 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ----------------------------------------------------------------------------------
+
+$envFile = "env.json"
+$env = @{}
+if(Test-Path -Path (Join-Path $PSScriptRoot $envFile)){
+    $env = Get-Content (Join-Path $PSScriptRoot $envFile) | ConvertFrom-Json
+} else {
+    $env.MsGraphTenantIdentifier = ${env:MSGRAPHTENANTIDENTIFIER}
```
MIchaelMainer, line 20: Consider raising an error in case these required environment variables aren't present. Error message should provide clear instructions on what variables need to be set in the environment.

scripts/TenantSetup/Install-Tools.ps1
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ----------------------------------------------------------------------------------
+
+function Install-MicrosoftGraph(){
+    if(-not (Get-Module Microsoft.Graph -ListAvailable)){
+        Install-Module Microsoft.Graph -Force
```
baywet, line 17: maybe install in the current user scope instead? installing globally requires admin permissions and depending on the CI platform we might not have that privilege

scripts/TenantSetup/TenantHttpData/Users/POST/user.http
```diff
@@ -3,11 +3,11 @@ Content-type: application/json
...
   "accountEnabled": true,
-  "displayName": "Adele VanceTheThird",
+  "displayName": "Adele Vance",
   "mailNickname": "AdeleV",
-  "userPrincipalName": "FinalVance2@MsGraphRaptorBase.onmicrosoft.com",
+  "userPrincipalName": "Adele Vance@<tenant>.onmicrosoft.com",
   "passwordProfile" : {
     "forceChangePasswordNextSignIn": true,
-    "password": "xWwvJ]6NMw+bWH-d"
+    "password": "<password>"
```
zengin, line 11: Thanks for making this change. We can also consider adding logic to automatically replace these with values from the test environment.

Given that you already have an envFile, we can maybe use that.

scripts/TenantSetup/loadEnv.ps1
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+$envFile = "env.json"
+$env = @{}
+if (Test-Path -Path (Join-Path $PSScriptRoot $envFile)) {
+    $env = Get-Content (Join-Path $PSScriptRoot $envFile) | ConvertFrom-Json
+}
+else {
+    $env.MsGraphTenantIdentifier = ${env:MSGRAPHTENANTIDENTIFIER}
+    $env.MsGraphClientIdentifier = ${env:MSGRAPHCLIENTIDENTIFIER}
```
baywet, line 9: nit: use underscore for environment variable names to improve readability

baywet (changes requested):

scripts/TenantSetup/loadEnv.ps1
finsharp: Given that these variables will come from Azure Devops for example, where its common to state them as `$env:SOME_VARIABLE`, wouldn't be ideal to be consistent with that pattern?

baywet: I'm not sure I've seen a convention for ADO. For GitHub people largely use underscores for readability as far as I can tell. Again it's a small thing but I think it'd improve readability. And we can name the variables whatever we want in effects. QQ: were those variable names set before you started the PS work or did you create those?

zengin (approved):

baywet (approved):

PR #221, opened 2021-03-31
Add/remove known issues

zengin:
- Removed 3 known issues as the root causes were fixed.
- Added more known issues. Deduplication introduced new names with numbers, it also allowed us to have more test cases, some of which are failing:
  - https://github.com/microsoftgraph/microsoft-graph-docs/pull/11975
- Marked consistently failing tests that need further analysis with `NeedsAnalysis`, so that we can follow PROD vs staging regressions. We will continue categorizing these in parallel.
- Added an explicit `Microsoft.Graph.Core` reference to prevent version conflict in the case of consuming an arbitrary dll, which may point to a never version than the published `Microsoft.Graph`.
- Reverted `Microsoft.Graph.Beta` version to latest preview that is earlier than `4.0.0-preview`. We are not ready to update to `4.0.0-preview` yet.
- Updated prerelease package updater to ignore `4.0.0-preview`.

Part of #196


scripts/PrereleasePackageUpdater.ps1
```diff
@@ -8,5 +8,30 @@ $projectFile = "msgraph-sdk-raptor-compiler-lib/msgraph-sdk-raptor-compiler-lib.
...
+    if ($package -eq "Microsoft.Graph.Beta")
+    {
+        $lowerCasePackage = $package.ToLower()
+        $nugetUrl = "https://api.nuget.org/v3/registration5-gz-semver2/$lowerCasePackage/index.json"
+        $res = Invoke-RestMethod $nugetUrl
+        $skipVersion = [version]"4.0.0"
+        $allVersionsLessThan4 = $res.items.items.catalogEntry.version | ForEach-Object {
+            if ($_.Contains("preview"))
+            {
+                $ver = [version]($_.Replace("-preview", ""))
```
MIchaelMainer, line 21: I think we are still good if we have a trailing `.1` like in `4.0.0-preview.1`. We don't use it in the beta version scheme.

MIchaelMainer (approved):

scripts/PrereleasePackageUpdater.ps1
zengin: yeah, this is a temporary fix anyways. Once we are back to normal, we can always continue getting the latest.

PR #236, opened 2021-04-08
JSON representation of the IDTree object

### LINKED ISSUE #235: JSON representation of the ID tree
Proposed JSON serialization of the `IDTree` object

If the tree is:
```
root
|
|-- application
|
|-- team
|   |
|   |-- channel
|   |   |
|   |   |-- conversationMember
|   |
|   |-- conversationMember
|
|-- chat
    |
    |-- conversationMember
```

Expected JSON would be:
```json
{
  "_value": null,
  "application": {
    "_value": "application-id"
  },
  "chat": {
    "_value": "chat-id",
    "conversationMember": {
      "_value": "chat-conversationMember-id"
    },
  },
  "team": {
    "_value": "team-id",
    "channel": {
      "_value": "team-channel-id",
      "conversationMember": {
        "_value": "team-channel-conversationMember-id"
      }
    }
  }
}
```

Key properties are:
1. Values are stored in `_value` property. Underscore is to prevent collision with a child item that may have `value` as its key.
2. Each child item appears as a property.
3. Child items are sorted within themselves for deterministic output (e.g. `chat` comes before `team` in the JSON) 
[AB#8898](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8898)

zengin:
Fixes #235 

- Implements custom JSON reader and writer
  - outputs `Value` into `_value` property
  - sorts keys in the dictionary of child items and outputs them recursively
- Implements IEquatable interface for comparing two `IDTree` objects.
- Adds equality and serialization tests.
- Adds a sample JSON file representing the serialized version of the ID tree extracted from V1 graph URLs.

### DIFF
--- a/TestsCommon/IDTree.cs
+++ b/TestsCommon/IDTree.cs
@@ -1,4 +1,8 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using System.Linq;
 
 namespace TestsCommon
 {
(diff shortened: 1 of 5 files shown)


TestsCommon/IDTree.cs
```diff
@@ -22,10 +26,77 @@ namespace TestsCommon
...
+    [JsonConverter(typeof(IDTreeConverter))]
+    public class IDTree : Dictionary<string, IDTree>, IEquatable<IDTree>
     {
         public string Value { get; set; }
         public IDTree(string value) { Value = value; }
+
+        public bool Equals(IDTree other)
+        {
+            return other != null &&
+                Value == other.Value &&
```
baywet, line 38: is case sensitivity important in the comparison?

TestsCommon/IDTree.cs
```diff
@@ -22,10 +26,77 @@ namespace TestsCommon
...
+                {
+                    return tree;
+                }
+
+                if (reader.TokenType != JsonTokenType.PropertyName)
+                {
+                    break;
+                }
+
+                var propertyName = reader.GetString();
```
baywet, line 71: java gson will throw if you call getString on a prop that's not a string. You need to test with isString first. I'm not sure whether system.text.json behaves the same way but making the suggestion just in case.

TestsCommon/IDTree.cs
```diff
@@ -22,10 +26,77 @@ namespace TestsCommon
...
+                if (propertyName == "_value")
```
baywet, line 72: CI comparison? or are all property names normalized?

TestsCommon/IDTree.cs
```diff
@@ -22,10 +26,77 @@ namespace TestsCommon
...
+                    tree = new IDTree(reader.GetString());
+                }
+                else
+                {
+                    IDTree subTree = JsonSerializer.Deserialize<IDTree>(ref reader, options);
+                    tree[propertyName] = subTree;
+                }
+            }
+
+            throw new JsonException();
```
baywet, line 84: maybe add a message to help differentiate with the defensive clause above?

baywet (changes requested):

TestsCommon/IDTree.cs
zengin: This is a case where the JSON would be invalid. I am OK with throwing in that scenario.

TestsCommon/IDTree.cs
zengin: They are normalized.

TestsCommon/IDTree.cs
zengin: We want an exact match in this case.

TestsCommon/IDTree.cs
```diff
@@ -22,10 +26,77 @@ namespace TestsCommon
     ///                                  |
     ///                             conversationMember
     /// </summary>
-    public class IDTree : Dictionary<string, IDTree>
+    [JsonConverter(typeof(IDTreeConverter))]
+    public class IDTree : Dictionary<string, IDTree>, IEquatable<IDTree>
     {
         public string Value { get; set; }
         public IDTree(string value) { Value = value; }
```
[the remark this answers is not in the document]
zengin: I initially designed the root node to have a null value, but skipping the property makes more sense. I will add a default constructor for that case and check for null here.

TestsCommon/IDTree.cs
```diff
@@ -22,10 +26,77 @@ namespace TestsCommon
...
+
+        public bool Equals(IDTree other)
+        {
+            return other != null &&
+                Value == other.Value &&
+                Keys.Count == other.Keys.Count &&
+                Keys.All(key => other.ContainsKey(key) && this[key].Equals(other[key]));
```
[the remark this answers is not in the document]
zengin: Thanks, added a test case as well.

baywet (approved):

PR #253, opened 2021-04-30
Request default token if we can't fetch the scopes needed from DevX API permissions endpoint

### LINKED ISSUE #252: DevX API doesn't return scopes for all URLs
We shouldn't fail if DevX API doesn't return the scopes for execution tests. We should continue attempting to run the test with default token. 
[AB#9236](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/9236)

zengin:
Fixes #252 

|   | Before | After
| - | - | - |
| Beta | 141/971 (14.5%) | 163/971 (16.8%) |
| V1 | 92/601 (15.3%)| 116/601 (19.3%) |

### DIFF
--- a/msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
+++ b/msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
@@ -197,16 +197,24 @@ namespace MsGraphSDKSnippetsCompiler
             using var scopesRequest = new HttpRequestMessage(HttpMethod.Get, $"https://graphexplorerapi.azurewebsites.net/permissions?requesturl={path}&method={httpRequestMessage.Method}");
             scopesRequest.Headers.Add("Accept-Language", "en-US");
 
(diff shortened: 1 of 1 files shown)


msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -197,16 +197,24 @@ static async Task<string[]> GetScopes(HttpRequestMessage httpRequestMessage)
...
+                    .ToList()
+                    .Where(x => x.value.Contains("Read") && !x.value.Contains("Write"))
+                    .Select(x => $"https://graph.microsoft.com/{ x.value }")
+                    .ToArray();
+
+                return fullReadScopes.Length == 0 ? null : fullReadScopes;
+            }
+            catch (Exception)
+            {
+                // some URLs don't return scopes from the permissions endpoint of DevX API
```
nikithauc, line 215: It would be nice to state the reason to return null which is mentioned in the issue.

nikithauc (approved):

PR #270, opened 2021-05-29
Simple token cache implementation

### LINKED ISSUE #269: Handle Client Request Loop Error
We get the following error when we run the tests from the same process. Implement caching for delegated permissions.

AADSTS50196:
LoopDetected - A client loop has been detected. Check the app’s logic to ensure that token caching is implemented, and that error conditions are handled correctly. The app has made too many of the same request in too short a period, indicating that it is in a faulty state or is abusively requesting tokens. 
[AB#9639](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/9639)

zengin:
Fixes #269
- Given that we are using the same user, the cache keys are scopes joined as a string.
- Since we acquire a lock while requesting a token, I removed the asynchronous call pattern.
- Ideally, we could make the network call outside the lock, but that complicates the design because we will then need a "processing" state for each cache key and two threads requesting the same key should synchronize separately. Because there isn't a significant perf impact of this change at the moment, I defer that exercise to a later time.

### DIFF
--- a/msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
+++ b/msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
@@ -20,7 +20,7 @@ using System.Reflection;
 using System.Text.RegularExpressions;
 using System.Security;
 using System.Text.Json;
-using Microsoft.Extensions.Configuration;
+using System.Collections.Concurrent;
 
 namespace MsGraphSDKSnippetsCompiler
 {
@@ -36,6 +36,10 @@ namespace MsGraphSDKSnippetsCompiler
         private const string AuthHeaderPattern = "Authorization: Bearer .*";
         private const string AuthHeaderReplacement = "Authorization: Bearer <token>";
         private static readonly Regex AuthHeaderRegex = new Regex(AuthHeaderPattern, RegexOptions.Compiled);
+
+        // token cache
+        private static readonly object tokenLock = new object();
+        private static ConcurrentDictionary<string, string> tokenCache = new ConcurrentDictionary<string, string>();
 
         private const string DefaultAuthScope = "https://graph.microsoft.com/.default";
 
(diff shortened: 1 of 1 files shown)


msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -142,7 +146,7 @@ public async Task<ExecutionResultsModel> ExecuteSnippet(string codeSnippet, Vers
                         var authority = config.GetNonEmptyValue("Authority");
                         var username = config.GetNonEmptyValue("Username");
                         var password = config.GetNonEmptyValue("Password");
-                        var token = await GetATokenForGraph(clientId, authority, username, password, scopes).ConfigureAwait(false);
+                        var token = GetATokenForGraph(clientId, authority, username, password, scopes);
```
baywet, line 149: what do you think about moving this into the body of the delegate to avoid long running operation in a ctor?

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -226,25 +230,39 @@ static async Task<string[]> GetScopes(HttpRequestMessage httpRequestMessage)
         /// <param name="password">password of the user for which the token is requested</param>
         /// <param name="scopes">requested scopes in the token</param>
         /// <returns>token for the given context</returns>
-        static async Task<string> GetATokenForGraph(string clientId, string authority, string username, string password, string[] scopes)
+        static string GetATokenForGraph(string clientId, string authority, string username, string password, string[] scopes)
         {
-            var app = PublicClientApplicationBuilder.Create(clientId).WithAuthority(authority).Build();
+            lock(tokenLock)
```
baywet, line 235: could the lock be at the scope below (the else)? I don't think we need to lock when getting the value.
Of course then the add would have to be done with tryadd instead, to avoid the case where thread A didn't find it and is trying to add it, but in the meantime thread B adds it.

zengin: Thought about it, but I would like to avoid two threads making the same requests, which causes the service to throw the throttling error. Given that the network operation is much slower than the shared memory access, having a larger lock scope doesn't add too much load.

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
zengin: I am not sure I follow what you refer to as "delegate" and "ctor" here.

baywet: ctor => constructor, delegate => the lambda being passed to DelegateAuthenticationProvider

zengin: thanks for clarifying. I was looking for those in the outer scope instead of the next line :D

MIchaelMainer (approved):

Issue #251, opened 2021-04-29
Extend data generation scripts to delegated permissions


zengin:
- For now, /me and /users/<user-id> APIs can be categorized to use delegated access.
- We need a step to fetch permission list from devx api, fetch a token, and use that in `Connect-MgGraph`. It needs to happen once per test, because scopes change from one test to another. 
[AB#9218](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/9218)

fey101:
Closed when the combined effort work was merged in PR https://github.com/microsoftgraph/msgraph-sdk-raptor/pull/272

PR #272, opened 2021-06-02
Read demo tenant and update identifiers.json

### LINKED ISSUE #255: Planner APIs require delegated permissions
Update delegated permissions check to include planner APIs. 
[AB#9311](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/9311)

### LINKED ISSUE #271: Create scripts to query existing data in the default data pack
Create scripts to query the data from demo tenant and store the results in an identifiers.json file. 
[AB#9668](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/9668)

zengin:
Fixes #271 
Fixes #255

- Creates two Powershell scripts that query existing data in the default data pack.
- Adds the template identifier replacement JSON file
- In local run, the template identifier JSON file is used so that local testing can skip relying on the blob storage.
- Adds more APIs that require delegated permissions. (#164 still needs to be addressed separately) 
- Data that need to be explored are left out as comments intentionally to guide us for pushing those missing data in the next phase of development.

### DIFF
--- a/scripts/readDefaultData.ps1
+++ b/scripts/readDefaultData.ps1
@@ -1,0 +1,264 @@
+$appSettingsPath = Join-Path $PSScriptRoot "../msgraph-sdk-raptor-compiler-lib/appsettings.json"
+$appSettings = Get-Content $appSettingsPath | ConvertFrom-Json
+
+if ( !$appSettings.CertificateThumbprint -or !$appSettings.ClientID -or !$appSettings.Username -or !$appSettings.TenantID)
+{
+    Write-Error -ErrorAction Stop -Message "please provide CertificateThumbprint, ClientID, Username and TenantID in appsettings.json"
+}
+
+$identifiersPath = Join-Path $PSScriptRoot "../msgraph-sdk-raptor-compiler-lib/identifiers.json"
+$identifiers = Get-Content $identifiersPath | ConvertFrom-Json
+
+$admin = "MOD Administrator"
+$domain = $appSettings.Username.Split("@")[1]
+$identifiers.domain._value = $domain
+
+function req
+{
+    param(
+        [string]$version = "v1.0",
+        [PSCustomObject]$headers = @{},
+        [string]$url
+    )
+
+    $response = Invoke-MgGraphRequest -Headers $headers -Method GET -Uri "https://graph.microsoft.com/$version/$url" -OutputType PSObject
+    $response.value
+}
+
+Connect-MgGraph -CertificateThumbprint $appSettings.CertificateThumbprint -ClientId $appSettings.ClientID -TenantId $appSettings.TenantID
+
+$user = req -url "users"  |
+    Where-Object { $_.displayName -eq $admin}
+    Select-Object -First 1
+$user.id
+$identifiers.user._value = $user.id
+
+$team = req -url "groups" |
+    Where-Object { $_.resourceProvisioningOptions -eq "Team" -and $_.displayName -eq "U.S. Sales"} |
+    Select-Object -First 1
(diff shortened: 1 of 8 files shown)


scripts/readDefaultData.ps1
```diff
@@ -0,0 +1,403 @@
...
+$admin = "MOD Administrator"
+$domain = $appSettings.Username.Split("@")[1]
+$identifiers.domain._value = $domain
+
+function req
+{
+    param(
+        [string]$version = "v1.0",
+        [PSCustomObject]$headers = @{},
+        [string]$url
```
fey101, line 29: We could reorder params such that required params comes first then default params.

scripts/readDefaultData.ps1
```diff
@@ -0,0 +1,403 @@
...
+    Where-Object { $_.displayName -eq "Global Administrator" }
+    Select-Object -First 1
+$directoryRole.id
+$identifiers.directoryRole._value = $directoryRole.id
+
+$directoryRoleTemplate = req -url "directoryRoleTemplates" |
+    Where-Object { $_.displayName -eq "Global Administrator" }
+    Select-Object -First 1
+$directoryRoleTemplate.id
+$identifiers.directoryRole._value = $directoryRoleTemplate.id
```
fey101, line 140: Did you intend to set the directoryRoleTemplate here instead of directoryRole? This would otherwise overwrite the value in line 134 above.

scripts/readDefaultData.ps1
```diff
@@ -0,0 +1,403 @@
...
+
+function req
+{
+    param(
+        [string]$version = "v1.0",
+        [PSCustomObject]$headers = @{},
+        [string]$url
+    )
+
+    $response = Invoke-MgGraphRequest -Headers $headers -Method GET -Uri "https://graph.microsoft.com/$version/$url" -OutputType PSObject
```
fey101, line 32: What happens if request fails due to network issues? could we provide a fail safe for this case?

scripts/readDefaultDataDelegated.ps1
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    $token.access_token
+}
+
+function reqDelegated
+{
+    param(
+        [string]$version = "v1.0",
+        [string]$url,
```
fey101, line 76: Order of param such that required param comes first.

fey101 (changes requested):
Just a few comments on order of params and update for directoryRoleTemplate

scripts/readDefaultData.ps1
zengin: We can rerun the full script in that case because rerunning the full script shouldn't affect the end result. In other words if we fail to update <type-id>, than <type-id> can be updated in a subsequent run.

scripts/readDefaultData.ps1
zengin: nice catch, thanks!

scripts/readDefaultData.ps1
zengin: thanks, updated.

scripts/readDefaultDataDelegated.ps1
zengin: thanks, updated.

georgend (approved):

fey101 (approved):

fey101:
Closes https://github.com/microsoftgraph/msgraph-sdk-raptor/issues/251

PR #276, opened 2021-06-07
Presence, TeamsApp and Minor Fixes. 

georgend:
Handle cases where returned data is an object.
Add ability to specify `-Debug` for debugging purposes. 
Add data fetch for:-
- Presence
- TeamsApp


fey101 (approved):
Thanks

PR #279, opened 2021-06-10
Add data Generation Scripts for entities under Teams

georgend:
Added Scripts to fetch or create the following entities.
- OfferShiftRequest
- OpenShift
- Schedule
- ScheduleShare
- SchedulingGroup
- Shift
- SwapShiftsChangeRequest
- TimeOffReason
- TimesOff


baywet:
This approach is going to require a lot of manual work if we want to provision a whole tenant this way. I just want to make sure that pnp provisioning templates we considered before we go too far. One of the benefits of those is they allow you to declare dependencies, i.e.: if you create (or ensure) a use, you can leverage the user id in a subsequent step (adding to a group).
https://docs.microsoft.com/en-us/sharepoint/dev/solution-guidance/pnp-provisioning-tenant-templates

An alternative (mostly for the configuration data) would be the Microsoft 365 DSC package, which allows to describe and apply configurations.
https://github.com/microsoft/Microsoft365DSC

PR #280, opened 2021-06-10
Add Script and Data for Policy Related Entities

georgend:
Adds Scripts and data to Populate Tenant with the following data if doesn't exist/
- ActivityBasedTimeoutPolicy
- ClaimsMappingPolicy
- HomeRealDiscoveryPolicy
- TokenIssuancePolicy


scripts/TenantSetup/TenantHttpData/Policy/Policy.ps1
```diff
@@ -0,0 +1,123 @@
...
+   Entity
+        - ChildEntity.json
+        - ChildEntity.json
+    Such as:-
+    Team
+        - OpenShift.json
+        - Schedule.json
+    Based on the Tree Structure in Identifiers.json
+#>
+function Get-RequestData {
```
baywet, line 29: these methods are duplicated with #279 , please move that to a another file and reference in both

baywet, line 29: these methods are duplicated with #279 , please move that to a another file and reference in both

baywet (changes requested):

scripts/TenantSetup/TenantHttpData/Policy/Policy.ps1
georgend: Noted, we are handling this.

baywet (approved):

fey101 (approved):

PR #286, opened 2021-06-15
Handle Verbatim Identifiers such as @event

georgend:
Currently Raptor incorrectly handles verbatim identifiers due to the Regex used to handle variable replacement. 

image

**The correct structure should be:-**

image

This PR changes the regex to correctly handle `@event`, `_event` and `@_event`.

**Other Changes**
- Check if Regex passes when replacing variables in C# test runner, throws and reports the error.
- Add tests to verify various cases for verbatim identifiers.

### DIFF
--- a/TestsCommon/CSharpTestRunner.cs
+++ b/TestsCommon/CSharpTestRunner.cs
@@ -156,15 +156,7 @@ public class GraphSDKTest
             string resultVariable = null;
             try
             {
-                var resultVariableMatch = ResultVariableRegex.Match(codeToCompile);
-                if (resultVariableMatch.Success)
-                {
-                    resultVariable = resultVariableMatch.Groups[1].Value;
-                }
-                else
-                {
-                    Assert.Fail("Regex {0}, against code {1} Failed", ResultVariablePattern, codeToCompile);
-                }
+                resultVariable = GetResultVariable(codeToCompile);
             }
             catch (Exception e)
             {
@@ -200,15 +192,7 @@ public class GraphSDKTest
             string resultVariable = null;
             try
             {
-                var resultVariableMatch = ResultVariableRegex.Match(codeSnippet);
-                if (resultVariableMatch.Success)
-                {
-                    resultVariable = resultVariableMatch.Groups[1].Value;
-                }
-                else
-                {
-                    Assert.Fail("Regex {0}, against code {1} Failed", ResultVariablePattern, codeSnippet);
-                }
+                resultVariable = GetResultVariable(codeSnippet);
             }
             catch (Exception e)
             {
@@ -221,6 +205,21 @@ public class GraphSDKTest
         return {resultVariable};");
 
             return codeSnippet;
+        }
+        private static string GetResultVariable(string codeToCompile)
+        {
+            string resultVariable = null;
+            var resultVariableMatch = ResultVariableRegex.Match(codeToCompile);
+            if (resultVariableMatch.Success)
+            {
+                resultVariable = resultVariableMatch.Groups[1].Value;
+            }
+            else
+            {
+                Assert.Fail("Regex {0}, against code {1} Failed", ResultVariablePattern, codeToCompile);
+            }
+
+            return resultVariable;
         }
 
         /// <summary>
--- a/TestsCommon/CSharpTestRunner.cs
+++ b/TestsCommon/CSharpTestRunner.cs
@@ -59,12 +59,12 @@ public class GraphSDKTest
         /// <summary>
         /// matches result variable name from code snippets
         /// </summary>
-        private const string ResultVariablePattern = "var ([a-zA-Z0-9]+) = await graphClient";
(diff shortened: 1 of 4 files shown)


TestsCommon/CSharpTestRunner.cs
```diff
@@ -151,12 +151,20 @@ private static void EvaluateCompilationResult(CompilationResultsModel compilatio
...
         }
 
-        private static string CaptureUriAndHeadersInException(string codeToCompile)
+        internal static string CaptureUriAndHeadersInException(string codeToCompile)
         {
             string resultVariable = null;
             try
             {
-                resultVariable = ResultVariableRegex.Match(codeToCompile).Groups[1].Value;
+                var resultVariableMatch = ResultVariableRegex.Match(codeToCompile);
```
baywet, line 159: refactor to a method with the one below to avoid duplication

TestsCommon/CSharpTestRunner.cs
```diff
@@ -151,12 +151,20 @@ private static void EvaluateCompilationResult(CompilationResultsModel compilatio
...
+                if (resultVariableMatch.Success)
+                {
+                    resultVariable = resultVariableMatch.Groups[1].Value;
+                }
+                else
+                {
+                    Assert.Fail("Regex {0}, against code {1} Failed", ResultVariablePattern, codeToCompile);
```
baywet, line 166: this is going to be caught by the catch and wrapper, was it intended this way? if so why not using a plain exception?

baywet (changes requested):

UnitTests/CSharpTestRunnerTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            Assert.AreEqual(match.Groups.Count, 2);
+            Assert.AreEqual(match.Groups[1].Value, variableName);
+        }
+
+        [TestCase("@event", VerbatimIdentifer, TestName = "@event")]
+        [TestCase("@_event", VerbatimIdentifierWithUnderScore, TestName = "@_event")]
+        [TestCase("_event", UnderScoreIdentifier, TestName = "_event")]
+        public void ShouldCaptureUriAndHeadersInException(string variableName, string testSnippet)
+        {
+            CSharpTestRunner.CaptureUriAndHeadersInException(testSnippet);
```
fey101, line 68: Are we asserting anything in this test "ShouldCaptureUriAndHeadersInException"

UnitTests/CSharpTestRunnerTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        [TestCase("@event", VerbatimIdentifer, TestName = "@event")]
+        [TestCase("@_event", VerbatimIdentifierWithUnderScore, TestName = "@_event")]
+        [TestCase("_event", UnderScoreIdentifier, TestName = "_event")]
+        public void ShouldReturnHttpRequestMessage(string variableName, string testSnippet)
+        {
+            CSharpTestRunner.ReturnHttpRequestMessage(testSnippet);
```
fey101, line 76: We should assert that it returns httpRequestMessage in some way, the first line of the returned string or any other portion of it.

TestsCommon/CSharpTestRunner.cs
```diff
@@ -59,12 +59,12 @@ public HttpRequestMessage GetRequestMessage(IAuthenticationProvider authProvider
...
         /// matches result variable name from code snippets
         /// </summary>
-        private const string ResultVariablePattern = "var ([a-zA-Z0-9]+) = await graphClient";
+        private const string ResultVariablePattern = "var ([@_a-zA-Z][_a-zA-Z0-9]+) = await graphClient";
 
         /// <summary>
         /// compiled version of the regex matching result variable name from code snippets
         /// </summary>
-        private static readonly Regex ResultVariableRegex = new Regex(ResultVariablePattern, RegexOptions.Singleline | RegexOptions.Compiled);
+        internal static readonly Regex ResultVariableRegex = new Regex(ResultVariablePattern, RegexOptions.Singleline | RegexOptions.Compiled);
```
fey101, line 67: 🥇 Am happy to see the usecase for internal access modifier. Learnt something new.

fey101 (changes requested):
LGTM. Just a few comments on some tests that lack assertions. Otherwise, well done.

TestsCommon/CSharpTestRunner.cs
fey101: I think it's intended so that it doesn't stop the rest of the raptor e2e tests from executing owing to the way the tests are structured such that it's basically a single test encompassing the tests for all the available code snippets. And given that raptor is generally a test case in it's structure maybe assert.Fail is the right approach.

UnitTests/CSharpTestRunnerTests.cs
georgend: I will use `Assert.DoesNotThrow()` to confirm that no exception is thrown by the call.

baywet (approved):

fey101 (approved):

Issue #285, opened 2021-06-14
Fill Missing data : Chat, DataPolicyOperations,OwnedDevices & ThreatAssessmentRequests


fey101:
[AB#9885](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/9885)

fey101:
Deferred Owned Devices for later.

PR #289, opened 2021-06-17
Fey 101/fill missing data part 1

fey101:
Closes https://github.com/microsoftgraph/msgraph-sdk-raptor/issues/285
- Rearrage functions in `RaptorUtils.ps1` so that the get app settings related functions move to the top with almost no change made to functions themselves. The one change made is the default $AppSettingsPath relative path. Tested this to ensure no breakage with previous scripts which explicitly set their paths.
- Added `reqDelegated` and `getToken` functions to `RaptorUtils.ps1` file
- Add identifiers.json and appsettings.json to gitignored files. There could be need to run the following command in local to clear git-tracking for this files
```bash
git update-index --assume-unchanged .\msgraph-sdk-raptor-compiler-lib\appsettings.json
git update-index --assume-unchanged .\msgraph-sdk-raptor-compiler-lib\identifiers.json
```
- Adds scripts to fill in missing data for 
     - chat
         - chat
         - conversationMember
         - teamsTab
         - installedApps
     - dataPolicyOperations
     - threatAssessmentRequest


scripts/TenantSetup/RaptorUtils.ps1
```diff
@@ -48,30 +78,79 @@ function Invoke-RequestHelper (
...
-    [string] $AppSettingsPath = (Join-Path $MyInvocation.PSScriptRoot "../../../../msgraph-sdk-raptor-compiler-lib/appsettings.json")
-) {
-    $appSettings = Get-Content $AppSettingsPath -Raw | ConvertFrom-Json
-    if (    !$appSettings.CertificateThumbprint `
-            -or !$appSettings.ClientID `
-            -or !$appSettings.Username `
-            -or !$appSettings.Password `
-            -or !$appSettings.TenantID) {
-        Write-Error -ErrorAction Stop -Message "please provide CertificateThumbprint, ClientID, Username, Password and TenantID in appsettings.json"
+function getToken
```
baywet, line 81: naming conventions: get-token

scripts/TenantSetup/RaptorUtils.ps1
```diff
@@ -48,30 +78,79 @@ function Invoke-RequestHelper (
...
+{
+    param(
+        [string]$path,
+        [string]$scopeOverride,
+        [Parameter(Mandatory = $False)]
+        [ValidateSet("GET", "POST", "PUT", "PATCH", "DELETE")]
+        [string] $method = "GET"
+    )
+
+    $tokenEndpoint = "https://login.microsoftonline.com/$domain/oauth2/v2.0/token"
```
baywet, line 91: can't we use the ps sdk to do that for us instead of crafting the request manually? CC @peombwa

scripts/TenantSetup/RaptorUtils.ps1
```diff
@@ -48,30 +78,79 @@ function Invoke-RequestHelper (
...
 }
 
-function Get-CurrentDomain (
-    [PSObject]$AppSettings
-) {
-    $domain = $appSettings.Username.Split("@")[1]
-    return $domain
-}
\ No newline at end of file
+function reqDelegated
```
baywet, line 133: naming convnetions

baywet (changes requested):

scripts/TenantSetup/RaptorUtils.ps1
fey101: We cant because ps sdk does not support delegated permissions for an app. So to bypass the need for user interaction when requesting delegated permission, we craft it manually.

baywet: thanks for clarifying!

baywet (approved):

scripts/TenantSetup/RaptorUtils.ps1
```diff
@@ -48,30 +78,79 @@ function Invoke-RequestHelper (
...
     }
-    return $appSettings
-}
 
-function Get-CurrentIdentifiers (
-    [string] $IdentifiersPath = (Join-Path $MyInvocation.PSScriptRoot "../../../../msgraph-sdk-raptor-compiler-lib/identifiers.json")
-) {
-    $identifiers = Get-Content $IdentifiersPath -Raw | ConvertFrom-Json
-    return $identifiers
+    $body = "grant_type=$grantType&username=$($appSettings.Username)&password=$($appSettings.Password)&client_id=$($appSettings.ClientID)&scope=$($joinedScopeString)"
```
MIchaelMainer, line 121: Might as well hardcode grantType as password since username and password parameters are only used with password grant.

scripts/TenantSetup/RaptorUtils.ps1
```diff
@@ -48,30 +78,79 @@ function Invoke-RequestHelper (
...
+    $token = Invoke-RestMethod -Method Post -Uri $tokenEndpoint -Body $Body -ContentType 'application/x-www-form-urlencoded'
+
+    Write-Debug "== got token with the following scopes"
+    foreach ($scope in $token.scope.Split())
+    {
+        Write-Debug "    $scope"
+    }
+
+    $token.access_token
```
MIchaelMainer, line 130: Even though this works as the returned value, for consistency, I suggest a return statement.

`return $token.access_token`

MIchaelMainer (changes requested):
nit: add copyright headers to all files, add documentation as well.

MIchaelMainer (approved):

PR #295, opened 2021-06-22
IdentifierReplacer throwing `KeyNotFoundException`

georgend:
IdentifierReplacer throwing `KeyNotFoundException` for tests whose entity key is not in IDTree(Identifiers.json).
Handle case when Key is not found. 
Add test to verify scenario.

image

Add missing entities to identifiers tree to resolve `KeyNotFoundException`
- EducationClass
- EducationAssignment
- EducationSubmission
- educationSubmissionResource

### DIFF
--- a/msgraph-sdk-raptor-compiler-lib/Models/IdentifierReplacer.cs
+++ b/msgraph-sdk-raptor-compiler-lib/Models/IdentifierReplacer.cs
@@ -3,6 +3,7 @@ using System.IO;
 using System.Text;
 using System.Text.Json;
 using System.Text.RegularExpressions;
+
 using Azure.Storage.Blobs;
 
 namespace MsGraphSDKSnippetsCompiler.Models
@@ -14,7 +15,7 @@ namespace MsGraphSDKSnippetsCompiler.Models
         /// </summary>
         public static IdentifierReplacer Instance => lazy.Value;
 
-        private static readonly Lazy<IdentifierReplacer> lazy = new (() => new IdentifierReplacer());
+        private static readonly Lazy<IdentifierReplacer> lazy = new(() => new IdentifierReplacer());
 
         /// <summary>
         /// tree of IDs that appear in sample Graph URLs
@@ -82,8 +83,12 @@ namespace MsGraphSDKSnippetsCompiler.Models
                 var id = match.Groups[0].Value;     // e.g. {site-id}
                 var idType = match.Groups[1].Value; // e.g. extract site from {site-id}
 
-                currentIdNode = currentIdNode[idType];
-                input = input.Replace(id, currentIdNode.Value);
+                var exists = currentIdNode.TryGetValue(idType, out IDTree localTree);
+                if (exists && localTree != null)
+                {
+                    currentIdNode = localTree;
+                    input = input.Replace(id, currentIdNode.Value);
+                }
             }
 
             return input;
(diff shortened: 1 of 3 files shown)


msgraph-sdk-raptor-compiler-lib/Models/IdentifierReplacer.cs
```diff
@@ -82,8 +83,12 @@ public string ReplaceIds(string input)
                 var id = match.Groups[0].Value;     // e.g. {site-id}
                 var idType = match.Groups[1].Value; // e.g. extract site from {site-id}
 
-                currentIdNode = currentIdNode[idType];
-                input = input.Replace(id, currentIdNode.Value);
+                var exists = currentIdNode.TryGetValue(idType, out IDTree localTree);
+                if (exists && localTree != null)
```
baywet, line 87: isn't this condition redundant? wouldn't testing localTree for non null be sufficient?

msgraph-sdk-raptor-compiler-lib/Models/IdentifierReplacer.cs
```diff
@@ -82,8 +83,12 @@ public string ReplaceIds(string input)
                 var id = match.Groups[0].Value;     // e.g. {site-id}
                 var idType = match.Groups[1].Value; // e.g. extract site from {site-id}
 
-                currentIdNode = currentIdNode[idType];
-                input = input.Replace(id, currentIdNode.Value);
+                var exists = currentIdNode.TryGetValue(idType, out IDTree localTree);
```
fey101, line 86: @georgend , the `TryGetValue` method is it defined anywhere? It doesn't seem to be in the IDTree class, and you haven't added it either.

fey101: nvm. Seen it's a function of a dictionary.

fey101 (commented):
Looks good to me. Won't approve as there's a change that could be made based on ongoing conversation with baywet.

msgraph-sdk-raptor-compiler-lib/Models/IdentifierReplacer.cs
georgend: We can remove the null check. It was overkill.

msgraph-sdk-raptor-compiler-lib/Models/IdentifierReplacer.cs
```diff
@@ -82,8 +83,12 @@ public string ReplaceIds(string input)
...
+                if (exists)
```
baywet, line 87: to be more specific, I was advocating to keep only the non-null check.
The entry can exist in the dictionary but be set to null, which would fail the next iteration.
However if the value is not null, it obviously exists. Which makes the exists variable assignment and check redundant.

baywet:
```suggestion
                if (localTree != null)
```

msgraph-sdk-raptor-compiler-lib/Models/IdentifierReplacer.cs
baywet:
```suggestion
                currentIdNode.TryGetValue(idType, out IDTree localTree);
```

baywet (changes requested):

baywet (approved):

Issue #299, opened 2021-06-28
DriveItem, Subscription, EducationClass and EducationSchool. WorkForceIntegration


fey101:
Fill missing data for above resources 
[AB#10063](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/10063)

zengin:
This was closed before introducing educationSchool, I opened a new issue to address that: #609. @fey101, please share if you have any findings around if there are any blockers around creating an educationSchool object.

PR #300, opened 2021-06-30
Add subscription, educationClass and workforceIntegration  data

fey101:
Partly fulfills #299 

Items added:
- Subscription
- educationClass
    - educationClass
        - educationAssignment
            - educationSubmission
                - educationSubmissionResource 
                
- workforceIntegration

To add Workbook in as a driveItem and related data points.

### DIFF
--- a/scripts/TenantSetup/TenantHttpData/educationClass/educationClass.ps1
+++ b/scripts/TenantSetup/TenantHttpData/educationClass/educationClass.ps1
@@ -1,0 +1,46 @@
+# Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+Param(
+    [string] $IdentifiersPath = (Join-Path $PSScriptRoot "../../../../msgraph-sdk-raptor-compiler-lib/identifiers.json" -Resolve)
+)
+$raptorUtils = Join-Path $PSScriptRoot "../../RaptorUtils.ps1" -Resolve
+. $raptorUtils
+
+$identifiers = Get-CurrentIdentifiers -IdentifiersPath $IdentifiersPath
+$appSettings = Get-AppSettings
+
+#Connect To Microsoft Graph Using ClientId, TenantId and Certificate
+Connect-MgGraph -CertificateThumbprint $appSettings.CertificateThumbprint -ClientId $appSettings.EducationClientId -TenantId $appSettings.EducationTenantId
+
+
+if($null -eq $educationClass){
+    $educationClass = Invoke-RequestHelper -Uri "education/classes" -Method GET |
+        Where-Object { $_.displayName -eq "Physical Science" } |
+        Select-Object -First 1
+}
+$educationClass.id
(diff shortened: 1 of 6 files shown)


scripts/TenantSetup/TenantHttpData/educationClass/educationClass.ps1
```diff
@@ -0,0 +1,46 @@
...
+if($null -eq $educationClass){
+    $educationClass = Invoke-RequestHelper -Uri "education/classes" -Method GET |
+        Where-Object { $_.displayName -eq "Physical Science" } |
+        Select-Object -First 1
+}
+$educationClass.id
+$identifiers.educationClass._value = $educationClass.id
+
+if($null -eq $educationAssignment){
+    $educationAssignment = Invoke-RequestHelper -Uri "education/classes/$($educationClass.id)/assignments?`$filter=displayName eq 'Midterm'" -Method GET |
```
MIchaelMainer, line 24: I'm certain this has already been answered but I'm curious. Why do we build the URI to make the request like this instead of using the PowerShell  SDK?

scripts/TenantSetup/TenantHttpData/workforceIntegration/workforceIntegration.json
```diff
@@ -0,0 +1,11 @@
+{
+    "displayName": "DemoWorkforceIntegration",
+    "apiVersion": 1,
+    "encryption": {
+      "protocol": "sharedSecret",
+      "secret": "uh5ZB27wWZVdZ3ahGNUg6m2WAywCv9Z2uJWnEcwNkSwdjFBPM8Qqtrin4bcpDmjJ"
```
MIchaelMainer, line 6: Was this intentional? Do we need to invalidate this secret?

fey101: I randomly generated this string, and wasn't sure if it is critical in any way. But since it will be used for creating an integration, I will invalidate it and find another way of providing this value.

scripts/TenantSetup/TenantHttpData/educationClass/educationClass.ps1
fey101: Because some commands are only available in beta but we need the data in V1 as well. eg the educationClassAssignment.

fey101: Also just to keep the pace faster, we would prefer to avoid rabbit holes that come with experimenting for now and maybe later when the sdk is more stable, we could rewrite this into powershell sdk commands.

scripts/TenantSetup/TenantHttpData/workforceIntegration/workforceIntegration.json
MIchaelMainer: I'm not familiar with the API so I wasn't sure if this was an actual secret or a place holder.

MIchaelMainer (approved):

PR #301, opened 2021-07-01
When no scopes are returned default to default scope.

georgend:
Due to missing permissions, DevX sometimes does not return requested permissions. 
As such default to the default scope ("https://graph.microsoft.com/.default") whenever:-
- DevX is unreachable 
- DevX returns HTTP 500

This change unblocks a significant number of tests such as:-
- get-user-presences-csharp-V1-compiles

Yields a significant jump from **~47pc to ~54pc.**
***Previous Run***
image
***This Run***
image


MIchaelMainer:
Won't this will hide the missing permissions from us? How do we discover the missing permissions?

georgend:
Fair point.
An error reporting feature, similar to APIDoctors Issues system might be a good solution.

MIchaelMainer:
@georgend I'll trade you an approval to unblock you for a tracking feature issue so that we get a report on missing permissions. Deal? Link to this PR.

MIchaelMainer (approved):
@georgend I took the deal on your behalf 😄

PR #313, opened 2021-07-13
Add workbook as a driveItem

### LINKED ISSUE #299: DriveItem, Subscription, EducationClass and EducationSchool. WorkForceIntegration

### LINKED ISSUE #325: Missing Data: Printer
Related Items:
- PrinterConnector
- PrinterShare
- PrintOperation
- PrintService has missing ChildItem -> PrintServiceEndpoint
- PrintTaskDefinition and Children
- PrintUsageByPrinter
- PrintUsageByUser
[AB#10252](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/10252)

fey101:
- Adds driveItem and nested data points. Closes #299 
- Adds printer and nested data points. Closes #325 
- Adds `InputFilePath` as an optional parameter to function `Request-DelegatedResource` to support uploading of files
- Ensures that `response.value` is returned in function `Request-DelegatedResource` only when the value points to an array and avoids situation where a returned single object with property `value` is assumed to be an odata listview response.
- Replaces the `else` condition with `if` so as to capture situations where a GET method call requires a permission that is concatenated with the write option e.g. *.ReadWrite.all. As was, this scenario would return the value of $joinedScopeString as an empty string even though the permissions endpoint actually had returned some required scopes that happened to all have the write permissions chained to the read.


georgend (changes requested):

scripts/TenantSetup/RaptorUtils.ps1
```diff
@@ -160,19 +160,31 @@ function Request-DelegatedResource
...
+    }
     Write-Debug "== getting token for $($Uri) for method $($Method)"
 
     $token = Get-Token -Path "/$Uri" -ScopeOverride $ScopeOverride -Method $Method
     Connect-MgGraph -AccessToken $token | Out-Null
 
     $jsonBody = $Body | ConvertTo-Json -Depth 3
-    $response = Invoke-MgGraphRequest -Method $Method -Headers $Headers -Uri "https://graph.microsoft.com/$Version/$Uri" -Body $jsonBody -OutputType PSObject -ResponseHeadersVariable "respHeaderVar"
-    return $response.value ?? $response ?? $respHeaderVar
+    if ($FilePath) {
```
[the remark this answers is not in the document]
fey101: Added.

scripts/TenantSetup/RaptorUtils.ps1
```diff
@@ -156,19 +156,31 @@ function Request-DelegatedResource {
...
 
     $token = Get-Token -Path "/$Uri" -ScopeOverride $ScopeOverride -Method $Method
     Connect-MgGraph -AccessToken $token | Out-Null
 
     $jsonBody = $Body | ConvertTo-Json -Depth 3
-    $response = Invoke-MgGraphRequest -Method $Method -Headers $Headers -Uri "https://graph.microsoft.com/$Version/$Uri" -Body $jsonBody -OutputType PSObject -ResponseHeadersVariable "respHeaderVar"
-    return $response.value ?? $response ?? $respHeaderVar
+    if ($FilePath -and (Test-Path -Path $FilePath)) {
+        # provide -InputFilePath param instead of -Body param
+        $response = Invoke-MgGraphRequest -Method $Method -Headers $Headers -Uri "https://graph.microsoft.com/$Version/$Uri" -InputFilePath $FilePath -OutputType PSObject -ResponseHeadersVariable "respHeaderVar"
```
MIchaelMainer, line 176: nit: `$Uri` is actually a relative URI, consider renaming it `$RelativeUri`

scripts/TenantSetup/TenantHttpData/driveItem/driveItem.ps1
```diff
@@ -0,0 +1,135 @@
...
+    [string] $workbookFilePath = (Join-Path $PSScriptRoot "./raptorWorkbookDriveItem.xlsx" -Resolve),
+    [string] $IdentifiersPath = (Join-Path $PSScriptRoot "../../../../msgraph-sdk-raptor-compiler-lib/identifiers.json" -Resolve)
+)
+$raptorUtils = Join-Path $PSScriptRoot "../../RaptorUtils.ps1" -Resolve
+. $raptorUtils
+
+$identifiers = Get-CurrentIdentifiers -IdentifiersPath $IdentifiersPath
+
+
+if($null -eq $driveItem) {
```
MIchaelMainer, line 12: global comment

Is it possible that these variables could already be set in the PowerShell context where this script runs? If so, we could have some unexpected behavior. Would it make sense for us to raise a warning message if these variables are already set when this script runs?

zengin: I think there is some dependency on the order of executing these PowerShell scripts. It was initially only two files, one for application permissions and the other for delegated permissions and we were using values from one script execution in another. It looks like we are expanding that into multiple files here with further categorization and scaling that can be confusing.

This also points to an indeterministic behavior for the output. If both negative and positive cases of the `if` statement can be true in different executions, we are outputting two different `identifiers.json` files which is unexpected.

I think we should make the execution order more explicit either by putting everything into a single file or some other mechanism.

scripts/TenantSetup/TenantHttpData/printer/printer.ps1
```diff
@@ -0,0 +1,65 @@
...
+    [string] $IdentifiersPath = (Join-Path $PSScriptRoot "../../../../msgraph-sdk-raptor-compiler-lib/identifiers.json" -Resolve)
+)
+Install-Az
+$raptorUtils = Join-Path $PSScriptRoot "../../RaptorUtils.ps1" -Resolve
+. $raptorUtils
+
+$identifiers = Get-CurrentIdentifiers -IdentifiersPath $IdentifiersPath
+$appSettings = Get-AppSettings -AppSettingsPath $AppSettingsPath
+
+if($null -eq $printer) {
```
zengin, line 13: @MIchaelMainer's comment holds here too.

scripts/TenantSetup/TenantHttpData/printer/printer.ps1
```diff
@@ -0,0 +1,65 @@
...
+    $storedSecretContent = Get-AzKeyVaultSecret -VaultName $appSettings.AzureKeyVaultName -Name "csrData" -AsPlainText
```
zengin, line 14: Please add a comment here explaining what we are achieving by calling these APIs.

zengin (changes requested):

scripts/TenantSetup/RaptorUtils.ps1
```diff
@@ -156,19 +156,31 @@ function Request-DelegatedResource {
...
+    }
     Write-Debug "== getting token for $($Uri) for method $($Method)"
 
     $token = Get-Token -Path "/$Uri" -ScopeOverride $ScopeOverride -Method $Method
     Connect-MgGraph -AccessToken $token | Out-Null
 
     $jsonBody = $Body | ConvertTo-Json -Depth 3
-    $response = Invoke-MgGraphRequest -Method $Method -Headers $Headers -Uri "https://graph.microsoft.com/$Version/$Uri" -Body $jsonBody -OutputType PSObject -ResponseHeadersVariable "respHeaderVar"
-    return $response.value ?? $response ?? $respHeaderVar
+    if ($FilePath -and (Test-Path -Path $FilePath)) {
```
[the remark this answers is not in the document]
fey101: $FilePath is optional. Having Test-Path alone would make it mandatory.

scripts/TenantSetup/RaptorUtils.ps1
fey101: @MIchaelMainer , this would affect lots of files that all pass the parameter $Uri. Could I handle this in a subsequent PR to reduce number of moving parts? I could create an issue and handle it as a separate PR from this one.

scripts/TenantSetup/TenantHttpData/driveItem/driveItem.ps1
fey101: With how the scripts work, it's okay to have the variables already set as it means the data had already been worked on/created for that particular variable. And if this is the case, there's no harm in avoiding subsequent http calls and relying on the data that was already there.

fey101: @zengin , the idea is still the same, in that everything that depends on driveItem is now in a single script. Which means the order of executing dependant scripts is implied. If printer for example was to influence driveItem.operation like it was initially thought, we'd bundle the printer logic within the driveItem.ps1 script. The only difference is that the expansion into multiple files brought about some modularity so that It's even easier and faster to tell the related/dependant scripts.

scripts/TenantSetup/TenantHttpData/printer/printer.ps1
fey101: Fixed.

scripts/TenantSetup/TenantHttpData/driveItem/driveItem.ps1
fey101: Additionally, I have removed the if blocks so that if driveItem changes everything below it is executed as well.

scripts/TenantSetup/RaptorUtils.ps1
zengin: @fey101 Relative URIs qualify as URIs, they are a special kind though. I agree that we should clarify that in the name, but I think it can be postponed. Please create an issue so that we don't forget.

zengin (approved):

scripts/TenantSetup/RaptorUtils.ps1
fey101: https://github.com/microsoftgraph/msgraph-sdk-raptor/issues/365

georgend (approved):

PR #341, opened 2021-07-23

zengin:
This is an automated PR.


baywet:
closing as I don't thing we want to revert to the old beta package at this point. CC @andrueastman

andrueastman:
Thanks @baywet. 
Any chance we know how to stop these ones? I believe we had a similar one earlier at #339

baywet:
@andrueastman this seems to be the automation triggering those PRs if you want to look into it. https://github.com/microsoftgraph/msgraph-sdk-raptor/blob/dev/.github/workflows/prerelease-package-updater.yaml

PR #347, opened 2021-07-27

zengin:
This is an automated PR.


baywet:
@andrueastman we get those daily, can you have a look and see if you can [fix the workflow](https://github.com/microsoftgraph/msgraph-sdk-raptor/blob/dev/.github/workflows/prerelease-package-updater.yaml) please?

zengin:
Should be fixed here: https://github.com/microsoftgraph/msgraph-sdk-raptor/pull/348

PR #348, opened 2021-07-27
Revert "modify prerelease package updater to special case 4.0.0"

zengin:
.NET v4 release made the special case stale, so reverting that.

cc: @baywet @andrueastman


baywet (approved):

andrueastman:
Thanks for this @zengin

PR #393, opened 2021-08-19
Json Parsing Exception in IDTree when _value is a Number

georgend:
IDTree values can be integers or floating points numbers, assumption that all _value are strings no longer holds and would require `.ToString()` in PowerShell data generation scripts. 
Add tests to verify new IDTree behavior.


zengin:
Why don't we keep everything as string though, given that the values we are replacing are always strings?

Maybe we should fix this in the generation scripts e.g. while assigning, _value = "$id_value_from_query" to force assign a string value.

fey101 (approved):
LGTM. Thanks for this fix

fey101:
> Why don't we keep everything as string though, given that the values we are replacing are always strings?
> 
> Maybe we should fix this in the generation scripts e.g. while assigning, `_value = "$id_value_from_query"` to force assign a string value.

@zengin , isn't that option restraining? The logic remains in our head and it's subject to not being adhered to if someone else without this knowledge works on the scripts. Why not have the same logic in code instead?

fey101:
Looks like the auto-merge was on and I had not read @zengin  comment before approving. My apologies.

georgend:
Letme undo the PR, till we address the other points.

PR #394, opened 2021-08-20
Revert "Json Parsing Exception in IDTree when _value is a Number"

georgend:
Reverts microsoftgraph/msgraph-sdk-raptor#393


fey101 (approved):

zengin:
@georgend, I think your approach is more defensive. I agree with @fey101's point here: https://github.com/microsoftgraph/msgraph-sdk-raptor/pull/393#issuecomment-902551569

I'd suggest we keep your previous change and close this PR. Sorry for the additional work.

PR #416, opened 2021-09-03
Delegated permissions with one app per scope

### LINKED ISSUE #303: Report on missing permissions from DevX API
Instead of ignoring missing permissions, we should report on them so that we can communicate gaps in permissions and get the permission mapping updated.

https://github.com/microsoftgraph/msgraph-sdk-raptor/issues/302
https://github.com/microsoftgraph/msgraph-sdk-raptor/pull/301 
[AB#10121](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/10121)

### LINKED ISSUE #400: On-the-fly app generation for delegated permissions
At the time of tenant creation:
- Have a single application (`PermissionManager`) with `Application.ReadWrite.All` and `DelegatedPermissionGrant.ReadWrite.All` permissions.

At the time of execution for a snippet that requires delegated permission:
- Use `PermissionManager` to create an application: `DelegatedPermissionApp`
- Assign desired OAuth2 permission to `DelegatedPermissionApp` using `PermissionManager`.
- Create a token using `DelegatedPermissionApp`
- Execute the actual code snippet from docs
- Delete `DelegatedPermissionApp` using `PermissionManager`
- Permanently delete `DelegatedPermissionApp` using `PermissionManager`

I have a proof of concept code ready for this. The approach has the potential to extend into the tenant creation scripts where we need more delegated permissions.

zengin:
- It turned out that we get very slow runtime to create one application per scope for each test case as proposed in #400. Instead the same approach is used to pre-deploy one app per permission with display name: `DelegatedApp <scope>`, e.g. `DelegatedApp User.Read`. This will need to be maintained with a periodic task to pre-deploy if new permissions are added. (Fixes #400 with a slight modification).
- Test setup phase gets all the permission descriptions from the devx-content repo, acquires token for each scope type using the pre-deployed apps and caches those tokens for use in each test case.
- We used to rely on a handcrafted route list for deciding on when to use delegated permissions. We are changing that to:
  - Get all delegated scopes from DevX API permissions endpoint and succeed if at least one token from that permission list succeeds.
  - If there are no delegated scopes returned, get application permissions
  - If DevX API doesn't return any scope for both categories, we fail the test. This caused new known issue type: `MissingPermissionScope` with some new tests failing. This is a good regression, because we want to remove our workaround on delegated routes and we want to catch if DevX API can't return them (fixes #303 cc: @MIchaelMainer). We want to fix the problem at the root so that the whole ecosystem benefits (e.g. Graph Explorer can't get the list of necessary scopes for these URLs that are coming from documentation).
  - If either we don't get delegated scopes from DevX API or the delegated tokens fail, we attempt to run the test with application permissions (we have an app with all the application permissions granted).

So the summary of the success criterion for snippet execution is whether we can execute the snippet with at least one of the permissions listed in the docs. There is planning going on for validating individual permissions.

TODO:
Create issues for
- [x] Periodic task to update the pre-deployed delegated permission apps (#418)
- [x] Caching tokens for each permission type periodically and getting them faster in one go at test setup. Currently this step takes a couple of minutes. (#419)


msgraph-sdk-raptor-compiler-lib/PermissionManager.cs
```diff
@@ -0,0 +1,156 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Graph;
+using Microsoft.Graph.Auth;
+using Microsoft.Identity.Client;
+using MsGraphSDKSnippetsCompiler.Models;
+using NUnit.Framework;
+
+namespace MsGraphSDKSnippetsCompiler
+{
```
MIchaelMainer, line 15:
```
/// <summary>
/// A Graph client to get information about all of the Azure Applications that have the permissions used in the test cases.
/// </summary>
```

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -213,67 +268,36 @@ static async Task<string[]> GetScopes(HttpRequestMessage httpRequestMessage)
...
-        /// <param name="scopes">requested scopes in the token</param>
-        /// <returns>token for the given context</returns>
-        private async Task<string> GetATokenForGraph(string[] scopes)
-        {
-            IAccount account = null;
-            if (TokenCache.ContainsKey(_config.Username))
+            catch
             {
-                account = TokenCache[_config.Username];
+                TestContext.Out.WriteLine($"Can't get scopes for scopeType=Delegated, url={httpRequestMessage.RequestUri}");
```
MIchaelMainer, line 287: `scopeType=Delegated` should be `scopeType=DelegatedWork`

MIchaelMainer (approved):

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
zengin:
```suggestion
                TestContext.Out.WriteLine($"Can't get scopes for scopeType=DelegatedWork, url={httpRequestMessage.RequestUri}");
```

Issue #320, opened 2021-07-16
Missing Data: Call, CallRecords


georgend:
[AB#10247](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/10247)

zengin:
{callRecords.callRecord-id} test cases:

- [ ] callrecord-getpstncalls-csharp-V1-executes
- [ ] get-callrecord-expanded-csharp-V1-executes
- [ ] get-callrecord-csharp-V1-executes
- [ ] get-callrecord-sessions-expanded-csharp-V1-executes
- [ ] get-callrecord-sessions-csharp-V1-executes
- [ ] callrecord-getdirectroutingcalls-csharp-V1-executes

fey101:
Needs creation of new bot and adding capabilities as documented here https://docs.microsoft.com/en-us/microsoftteams/platform/bots/calls-and-meetings/registering-calling-bot

fey101:
It also requires the account to be authenticated via authenticator app.
Also when logging in with our test tenant I am warned with following error:
"[Teams Toolkit] Your Microsoft 365 account doesn't have a sideloading permission. Without this permission, you cannot load your app within Teams. Contact to your administrator to resolve this issue or signin another account with sideloading permission."

PR #421, opened 2021-09-09
Move csharp new known issues from stable tests

fey101:
Most of these tests are failing with the error 

"System.IO.InvalidDataException : no data found for id: {termStore.group-id} in identifiers.json file". 

For the above case, I know for example call records is still in our [todo issue list](https://github.com/microsoftgraph/msgraph-sdk-raptor/issues/320) data hasn't been created yet. Not sure how they were passing before. For these tests I have categorized them as "MissingDataKnownIssue"

A few others are failing with the error:

"Can't get scopes for both delegated and application scopes (The JSON value could not be converted to MsGraphSDKSnippetsCompiler.Models.Scope[]. Path: $ | LineNumber: 0 | BytePositionInLine: 1.)
The JSON value could not be converted to MsGraphSDKSnippetsCompiler.Models.Scope[]. Path: $ | LineNumber: 0 | BytePositionInLine: 1."

For the ones failing with can't get scopes, I have categorized as "MissingPermissionScopeKnownIssue"


baywet (approved):

zengin (changes requested):
Please mark the tests that can't get the scopes from DevX API as `MissingPermissionScopeKnownIssue`.

zengin (approved):

georgend (approved):

PR #439, opened 2021-09-15
Use Azure Identity instead of deprecated Microsoft.Graph.Auth

### LINKED ISSUE #307: Refactor authentication code to implement TokenCredentials
Version 4 of the .Net sdk will implement the use of `TokenCredentials` from `Azure.Identity`. 

It would be great if the raptor tests and code implemented TokenCredentials rather IAuthenticationProvider instances from the now deprecated Microsoft.Graph.Auth packages.

[AB#10158](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/10158)

zengin:
- removed references to Microsoft.Graph.Auth
- removed customized token cache
- started using TokenCredentialAuthProvider
- removed custom token access code
- removed tear down code that is no longer needed
fixes #307


msgraph-sdk-raptor-compiler-lib/PermissionManager.cs
```diff
@@ -276,45 +268,21 @@ internal async Task PopulateTokenCache()
                 try
                 {
                     var application = await GetApplication(delegatedPermissionScope);
-                    var token = await GetDelegatedAccessToken(application, scopeName);
-                    _tokenCache[delegatedPermissionScope.value] = token;
+                    _authProviders[delegatedPermissionScope.value] = new TokenCredentialAuthProvider(
+                        new UsernamePasswordCredential(_config.Username, _config.Password, _config.TenantID, application.AppId, new UsernamePasswordCredentialOptions
+                        {
+                            TokenCachePersistenceOptions = new TokenCachePersistenceOptions
```
andrueastman, line 274: 👍🏼

andrueastman (commented):
I like how things are simpler with the changes.

I suspect that there is a regression introduced with the changes though(execution tests seems to have lowered pass rate). So, we might want to investigate first before merging.

zengin:
@andrueastman interesting, thanks for raising that. There was no regression in the pass rate on Windows. Maybe the caching behavior is not working similarly on Linux. I will try to locally reproduce the issue.

zengin:
@andrueastman and thanks a lot again for the great work integrating Azure.Identity on v4. The auth code is much simpler now.

msgraph-sdk-raptor-compiler-lib/PermissionManager.cs
```diff
@@ -246,28 +239,27 @@ internal async Task<Application> GetApplication(Scope scope)
...
-        ///   2. b. Acquires a token using that preconfigured app
-        ///   2. c. Saves the result in the token cache for immediate use in the test.
+        ///   2. b. Creates a token credential provider for the app
         /// </summary>
         /// <returns></returns>
-        internal async Task PopulateTokenCache()
+        internal async Task CreateDelegatedAuthProviders()
         {
-            _tokenCache = new Dictionary<string, string>();
+            _authProviders = new Dictionary<string, TokenCredentialAuthProvider>();
```
baywet, line 262: maybe initialize the property directly and mark as readonly?

zengin: Initially I thought that maybe some initializations of this class didn't need the property, so I didn't want to initialize in the constructor. But it seems that we always need it, so I will move it up to constructor with `readonly`.

baywet (approved):

georgend (approved):

PR #443, opened 2021-09-16
App config refactoring

### LINKED ISSUE #441: Eliminate DocsRepoCheckoutDirectory setting from app config
### Discussed in https://github.com/microsoftgraph/msgraph-sdk-raptor/discussions/430

<div type='discussions-op-text'>

<sup>Originally posted by **zengin** September 14, 2021</sup>
Given that we have started using environment variables as a prerequisite to run Raptor tests, I think we can add `BUILD_SOURCESDIRECTORY` to that list. That will eliminate the need for having different file paths set for different developers in the app config. This is the only setting I need to differ for my Linux/Mac dev environments as well.

There is a way to set environment variables for codespaces too (it will be `/workspaces` because that is the standard source path).

Another advantage is that we will have a single code path for Azure DevOps runs and our local/codespace runs.

What do you think @fey101 @georgend?</div>

### LINKED ISSUE #442: Allow specifying custom labels for selecting app config from Azure
Currently we only have "Development" and "CI" labels. Allow customized labels for testing in local environments.

zengin:
- read docs directory from environment variable shared with Azure DevOps, fixes #441 
- singleton access to raptor config per execution
- deduce IsLocalRun from env variables (BUILD_REASON and BUILD_BUILDID are expected to be set only in Azure DevOps runs).
- allow custom app config labels for testing different settings locally, fixes #442 
- remove indexed access to raw config, reuse raptor config

### DIFF
--- a/msgraph-sdk-raptor-compiler-lib/Models/TestsSetup.cs
+++ b/msgraph-sdk-raptor-compiler-lib/Models/TestsSetup.cs
@@ -4,11 +4,21 @@ namespace MsGraphSDKSnippetsCompiler.Models
 {
     public static class TestsSetup
     {
+        private static RaptorConfig _raptorConfig;
+        private static readonly object _configLock = new { };
+
         public static RaptorConfig GetConfig()
         {
-            var config = AppSettings.Config();
-            var raptorConfig = RaptorConfig.Create(config);
-            return raptorConfig;
+            lock (_configLock)
+            {
+                if (_raptorConfig == null)
+                {
+                    var config = AppSettings.Config();
+                    _raptorConfig = RaptorConfig.Create(config);
+                }
+
+                return _raptorConfig;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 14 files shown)


msgraph-sdk-raptor-compiler-lib/Models/TestsSetup.cs
```diff
@@ -4,11 +4,21 @@ namespace MsGraphSDKSnippetsCompiler.Models
...
     {
+        private static RaptorConfig _raptorConfig;
+        private static readonly object _configLock = new { };
+
         public static RaptorConfig GetConfig()
         {
-            var config = AppSettings.Config();
-            var raptorConfig = RaptorConfig.Create(config);
-            return raptorConfig;
+            lock (_configLock)
```
baywet, line 12: out of curiosity, why not using a Lazy<RaptorConfig> instead?

zengin: Thanks for the reminder. No particular reason. I guess I was `Lazy<Mustafa>` to refactor further for all the calls :D Updated now.

baywet (approved):

PR #444, opened 2021-09-16

zengin:
This is an automated PR.


zengin:
Need to remove Microsoft.Graph.Auth from the prerelease updater.

PR #455, opened 2021-09-17
Make individual tests debuggable in VSCode

### LINKED ISSUE #454: Debuggability in VSCode
Make individual tests debuggable in VSCode. Consider setting up launch.json and tasks.json for projects and test filtering.

Couple of challenges:
- `dotnet test` can't be called in debug mode for attaching VSCode as the debugger. The workaround is using `VSTEST_HOST_DEBUG` environment variable.
- We generate test cases at runtime, OmniSharp's `debug test` invokes the full suite and doesn't allow individual test debugging as opposed to VS's Test Explorer.

zengin:
fixes #454

- Updated `tasks.json` to debug individual tests in VSCode and Github Codespaces. VSTEST_HOST_DEBUG allows attaching debugger to a `dotnet test` process. (See more details here: https://stackoverflow.com/questions/47066356/how-does-one-debug-an-mstest-in-vscode/47241515#47241515).
- Added tasks.json generator to reuse after extending the solution with more projects.
- Added a task to checkout docs repo as it is a prerequisite to run anything on Raptor.

TODO:
- [ ] #457

### DIFF
--- a/scripts/tasks/checkout-docs-repo.ps1
+++ b/scripts/tasks/checkout-docs-repo.ps1
@@ -1,0 +1,21 @@
+param(
+    [Parameter(Mandatory=$true)][string]$rootDirectory,
+    [string]$branchName = "main"
+)
+
+$docsRepoName = "microsoft-graph-docs"
+
+Set-Location $rootDirectory
+$docsRepoExists = Test-Path $docsRepoName
+if (!$docsRepoExists)
+{
+    git clone "https://github.com/microsoftgraph/$docsRepoName"
+}
+
+Set-Location $docsRepoName
+git fetch
+git reset --hard
+git checkout $branchName
+git pull -f
+
+Write-Host "checkout docs repo script"
--- a/scripts/tasks/checkout-docs-repo.ps1
+++ b/scripts/tasks/checkout-docs-repo.ps1
@@ -1,3 +1,10 @@
+# Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+
+<#
+.SYNOPSIS
+  This script checks out microsoft-graph-docs repo or resets to a particular branch
+#>
+
 param(
     [Parameter(Mandatory=$true)][string]$rootDirectory,
     [string]$branchName = "main"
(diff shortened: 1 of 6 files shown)


scripts/tasks/checkout-docs-repo.ps1
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
```
baywet, line 1: why don't we have the docs repo as a submodule of this one? it would make a lot of things easier here

zengin: Submodules are great but I don't think they make our life any easier in this particular repository.

- By keeping the docs repo separate, we can specify a branch directly in DevOps pipelines without any extra configuration.

- We can quickly test local changes in docs, without any effect on the working directory of Raptor.

- We don't always need the newest and latest from docs, but when we do (in daily and weekly pipelines), we get it for free without any dependabot maintenance. Note that docs/main gets updated several times a day.

scripts/tasks/checkout-docs-repo.ps1
```diff
@@ -0,0 +1,35 @@
...
+
+<#
+.SYNOPSIS
+  This script checks out microsoft-graph-docs repo or resets to a particular branch
+#>
+
+param(
+    [Parameter(Mandatory=$true)][string]$rootDirectory,
+    [string]$branchName = "main",
+    [string]$confirmation = "NO"
```
baywet, line 11: Have you considered a switch parameter instead?
https://social.technet.microsoft.com/wiki/contents/articles/23826.powershell-how-to-use-switch-parameter.aspx

zengin: Yes, this is fed by a text input from the command palette of VSCode, and there is no way to give an option list there. At least I couldn't find any. That's why I chose string type to give the warning message below if "YES" is not provided. Giving a parameter error could be confusing for someone just consuming the abstraction layer of VSCode tasks.

Display text of the input prompt indicates that positive response is YES and negative response is anything else.

baywet: oh ok, I missed the part where you ask for user input, I thought I was just command parameters (static list in the tasks)

baywet (approved):

PR #481, opened 2021-09-29
Make test output parsable for DevX API 404 responses for permissions

### LINKED ISSUE #480: Compile a list for DevX API 404 responses for permissions

zengin:
- Refactored the code to access link to documentation page (`DocsLink`) at exception time (first commit: https://github.com/microsoftgraph/msgraph-sdk-raptor/commit/d500543a721d0437a0c324d9e117b7afa514bad5).
  - Flattened `ExecutionTestData` and `LanguageTestData` types into a single `LanguageTestData` type because `LanguageTestData` already had execution related properties and `ExecutionTestData` was not adding too much value.
  - Made `MicrosoftGraphCSharpCompiler` type to refer to `LanguageTestData` directly so that we don't have to plumb properties one by one. We needed `DocsLink` in this particular occasion, but `FileName` and `DllPath` were already being plumbed.
- Tweaked the output to be parsable and added a quick PowerShell parser to create a table output for DevX API permission issues (second commit: https://github.com/microsoftgraph/msgraph-sdk-raptor/commit/845314622e91b0b274b7535248b2dee0e69d57c1)

fixes #480

### DIFF
--- a/TestsCommon/CSharpTestRunner.cs
+++ b/TestsCommon/CSharpTestRunner.cs
@@ -89,14 +89,10 @@ public class GraphSDKTest
                 throw new ArgumentNullException(nameof(testData));
             }
 
-            var fullPath = Path.Join(GraphDocsDirectory.GetSnippetsDirectory(testData.Version, Languages.CSharp), testData.FileName);
-            Assert.IsTrue(File.Exists(fullPath), "Snippet file referenced in documentation is not found!");
-
-            var fileContent = File.ReadAllText(fullPath);
-            var (codeToCompile, codeSnippetFormatted) = GetCodeToCompile(fileContent);
+            var (codeToCompile, codeSnippetFormatted) = GetCodeToCompile(testData.FileContent);
 
             // Compile Code
-            var microsoftGraphCSharpCompiler = new MicrosoftGraphCSharpCompiler(testData.FileName, testData.DllPath);
+            var microsoftGraphCSharpCompiler = new MicrosoftGraphCSharpCompiler(testData);
             var compilationResultsModel = microsoftGraphCSharpCompiler.CompileSnippet(codeToCompile, testData.Version);
 
             var compilationOutputMessage = new CompilationOutputMessage(compilationResultsModel, codeToCompile, testData.DocsLink, testData.KnownIssueMessage, testData.IsCompilationKnownIssue);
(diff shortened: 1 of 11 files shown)


TestsCommon/CSharpTestRunner.cs
```diff
@@ -112,20 +108,14 @@ public static void Compile(LanguageTestData testData)
...
         {
-            if (executionTestData == null)
-            {
-                throw new ArgumentNullException(nameof(executionTestData));
-            }
-
-            var testData = executionTestData.LanguageTestData;
 
-            var (codeToCompile, codeSnippetFormatted) = GetCodeToExecute(executionTestData.FileContent);
+            var (codeToCompile, codeSnippetFormatted) = GetCodeToExecute(testData.FileContent);
```
baywet, line 115: defensive programing on test data

msgraph-sdk-raptor-compiler-lib/MicrosoftGraphCSharpCompiler.cs
```diff
@@ -45,11 +44,9 @@ private async Task<PermissionManager> GetPermissionManager()
         private const string AuthHeaderReplacement = "Authorization: Bearer <token>";
         private static readonly Regex AuthHeaderRegex = new Regex(AuthHeaderPattern, RegexOptions.Compiled);
 
-        public MicrosoftGraphCSharpCompiler(string markdownFileName,
-            string dllPath)
+        public MicrosoftGraphCSharpCompiler(LanguageTestData testData)
         {
-            _markdownFileName = markdownFileName;
-            _dllPath = dllPath;
+            TestData = testData;
```
baywet, line 49: defensive programing on test data

baywet (changes requested):

TestsCommon/CSharpTestRunner.cs
baywet: unresolving, it feels this one was missed

baywet (approved):

TestsCommon/CSharpTestRunner.cs
zengin: ah sorry, yes. Thanks! I should setup SonarCloud ASAP ;)

baywet: we might get some odd results as we're generating code on the fly to compile and test it, that shapes our code a specific way, hopefully we don't get too many false positives here.

zengin: looking forward to this feature in C#

https://github.com/dotnet/csharplang/blob/main/proposals/null-arg-checking.md

It will save me a lot of lines :D

baywet: I wasn't aware of that spec, and yeah, it'll save tons of boilerplate code. Do we have a line of sight of when it'll lend?
I'm almost tempted to PR the spec and add !! which would not only check for null, but also for empty for collections 😂

zengin: dotnet/csharplang repo, issue 2145
(avoiding the full link here in order not to pollute the discussion with a reverse link to this repo)

PR #559, opened 2021-11-03
Remove incorrect orgContact Query, replace with query to users contacts.

### LINKED ISSUE #517: contact-id is not a contact of the admin
The following query fails:
```
https://graph.microsoft.com/v1.0/me/contacts/{contact-id}
```

Test Case: `get-contact-csharp-V1-executes`

georgend:
Replace with correct contact query.
Add User.ps1 and Contact data.

Closes #517


zengin (changes requested):
let's discuss this in one of our syncs.

zengin (approved):

Issue #660, opened 2021-11-17
Triage SDK issues found in Raptor for Kiota


zengin:
- [ ] https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1003
- [ ] https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/issues/471

baywet:
@zengin maybe we can handle most of that async. Is the question "is kiota going to address this issue?" for each line item?

zengin:
@baywet I am OK with async, I'd actually prefer that.

That's the main question, yes. When the validation is ready for kiota-generated SDK, we can double check with this list.

I think it would be nice to link Raptor test results directly to issues in the kiota repo. I would like to understand what that mapping looks like. I think there was a reason to keep some of these in the service library so that they are easy to find for the users of the SDKs. Maybe we can use a common label language to workaround that or some sort of 1:1 mapping of issues.

Thoughts?

baywet:
I'll leave the tagging and mapping of the issues in Raptor up to you, you know the area much better.

As for that issue list, would it work for you if I commented on each of them whether kiota already fixes that or work on creating issues in dedicated repos if work needs to happen.

zengin:
@baywet Sounds good to me. Please let me know when you are done, so that I will pick up the tagging and linking.

baywet:
I've left comments on all of them, don't hesitate to follow up if you have questions/remarks. This was useful as it also uncovered additional conversion process bugs.

On the checklist itself: 
-  https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/875 and https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/issues/402 are effectively the same thing, maybe update your mapping to only use one of those?
- https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1003 is duplicated in the checklist.

zengin:
Thanks for the quick action @baywet.

Tracking question (@darrelmiller @ddyett @MIchaelMainer @baywet @andrueastman)

Is it fair to assume that we will **not** fix any of the SDK issues in Typewriter anymore? If so, what should we do about issues that live in the generator repo?

For example, I would like to point to the OData OpenAPI conversion bug as the root cause if that is the only thing that is blocking generation of the $count url segment for our future SDK releases: https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/issues/402#issuecomment-971903236

[...]

PR #662, opened 2021-11-18
Known issue cleanup

zengin:
- Merge generator and service library issues regarding $count segment into one as suggested here: https://github.com/microsoftgraph/msgraph-sdk-raptor/issues/660#issuecomment-971916157
- remove unnecessary `HTTPMethodWrong` category given that we have only one instance remaining and merge it with `HTTP`.
- Re-root cause shift-put samples where the issue is in the additional data field for complex types.
- Add Github issue link to education school issue

partial #196


fey101 (approved):
LGTM

zengin:
the build passed, but for some reason Github has a problem between the Action and the PR to communicate.

PR #672, opened 2021-11-19
IdentiferReplacer to replace only first instance

### LINKED ISSUE #671: IdentifierReplacer calls string.Replace which replaces all instances
Depending on where the ID appears in a tree, different instances of the same ID might be required. So when there is a hit in the identifiers tree, we should only replace the first instance.

related: https://github.com/microsoftgraph/microsoft-graph-devx-api/issues/744

zengin:
fixes #671


msgraph-sdk-raptor-compiler-lib/Models/IdentifierReplacer.cs
```diff
@@ -131,10 +131,43 @@ private string ReplaceIdsFromIdentifiersFile(string input)
...
+    /// <param name="replacement">Replacement string</param>
+    /// <returns>input string with first instance of substring replaced</returns>
+    private static string ReplaceFirst(string input, string substring, string replacement)
+    {
+        if (input == null)
+        {
+            throw new ArgumentNullException(nameof(input));
+        }
+
+        if (substring == null)
```
baywet, line 155:
```suggestion
        if (string.IsNullOrEmpty(substring))
```

baywet (changes requested):

baywet (approved):

PR #678, opened 2021-11-24
add numbers into ID regex

### LINKED ISSUE #677: idRegex doesn't match IDs with numbers in them
We don't fail early with "missing data" message for the following URL, because ID regex is: `{([A-Za-z\.]+)\-id}`.

https://graph.microsoft.com/beta/identity/b2xUserFlows/{b2xIdentityUserFlow-id}/identityProviders

Test case affected: get-b2xuserflow-list-identityproviders-csharp-Beta-executes

zengin:
fixes #677

### DIFF
--- a/msgraph-sdk-raptor-compiler-lib/Models/IdentifierReplacer.cs
+++ b/msgraph-sdk-raptor-compiler-lib/Models/IdentifierReplacer.cs
@@ -20,7 +20,7 @@ public class IdentifierReplacer
     /// regular expression to match strings like {namespace.type-id}
     /// also extracts namespace.type part separately so that we can use it as a lookup key.
     /// </summary>
-    private readonly Regex idRegex = new Regex(@"{([A-Za-z\.]+)\-id}", RegexOptions.Compiled);
+    private readonly Regex idRegex = new Regex(@"{([A-Za-z0-9\.]+)\-id}", RegexOptions.Compiled);
 
     public IdentifierReplacer(IDTree tree)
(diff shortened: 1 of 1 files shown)


msgraph-sdk-raptor-compiler-lib/Models/IdentifierReplacer.cs
```diff
@@ -20,7 +20,7 @@ public class IdentifierReplacer
     /// regular expression to match strings like {namespace.type-id}
     /// also extracts namespace.type part separately so that we can use it as a lookup key.
     /// </summary>
-    private readonly Regex idRegex = new Regex(@"{([A-Za-z\.]+)\-id}", RegexOptions.Compiled);
+    private readonly Regex idRegex = new Regex(@"{([A-Za-z0-9\.]+)\-id}", RegexOptions.Compiled);
```
baywet, line 23:
```suggestion
    private readonly Regex idRegex = new Regex(@"{([\w\.]+)\-id}", RegexOptions.Compiled);

```

if underscore is ok as well.

zengin: Underscore is not allowed by AGS explicitly in their naming guideline.

https://msazure.visualstudio.com/One/_wiki/wikis/Microsoft%20Graph%20Partners/103140/Naming-guidelines?anchor=general-guidelines

baywet (approved):

PR #681, opened 2021-11-25
Add PSScriptAnalyzer and Pester to lint PS Scripts and display lint results. 

georgend:
Use [PSScriptAnalyzer](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-powershell#using-psscriptanalyzer-to-lint-code) to lint and report issues with the growing number of PowerShell Scripts.

Uses [Pester](https://devblogs.microsoft.com/scripting/psscriptanalyzer-deep-dive-part-3-of-4/) as described to format the results for easy consumption. 

Cleanup all scripts to remove flagged issues. 
Add suppressions for non-fixable issues and false positives. 

***Validation***
image

Lint PSScripts should pass (Green Check Mark)
And `0 Rule Violations` should be reported.


baywet (approved):

baywet (approved):

scripts/TenantSetup/RaptorUtils.ps1
```diff
@@ -1,6 +1,16 @@
-# Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
-
-function Get-AppSettings () {
+﻿# Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+
+[System.Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSAvoidUsingConvertToSecureStringWithPlainText', '', Scope='Function', Target='Get-HtmlDataRequest')]
```
fey101, line 3: This does not look neat. Is there a better way to do the same like having a linter-specific file where the unapplicable rules are suppressed?

scripts/TenantSetup/RaptorUtils.ps1
```diff
@@ -179,8 +189,6 @@ function Get-Scopes {
 #>
 function Get-ScopeString {
     param (
-        [Parameter(Mandatory = $False)][ValidateSet("GET", "POST", "PUT", "PATCH", "DELETE")][string] $Method = "GET",
```
fey101, line 182: Beauty of having linters that can quickly discover unused params. 👏🏽

scripts/generateCSharpModels.ps1
```diff
@@ -22,23 +23,23 @@ $cleanMetadataFile = Join-Path $metadataDir "clean_metadata.xml"
 
 # transform script uses relative path to working directory
 & $xslTranformScriptPath -xslPath $xslPath -inputPath $metadataFile -outputPath $cleanMetadataFile
-Write-Host "Transformed $metadataFile to $cleanMetadataFile" -ForegroundColor Green
+Write-Output "Transformed $metadataFile to $cleanMetadataFile" -ForegroundColor Green
```
fey101, line 26: with the Write-Host command, you can decorate the output with Foreground and Background color, and this is not possible with the Write-Output cmdlet.

scripts/TenantSetup/InitTenant.ps1
```diff
@@ -53,6 +53,6 @@ $nestedScripts | ForEach-Object {
     $scriptName = $_.Name
     $scriptPath = $_.FullName
 
-    Write-Host "Executing $scriptName at $scriptPath" -ForegroundColor Green
+    Write-Output "Executing $scriptName at $scriptPath" -ForegroundColor Green
```
fey101, line 56: with the Write-Host command, you can decorate the output with Foreground and Background color, and this is not possible with the Write-Output cmdlet. Any reason why you prefer write-output to write-host? I see "Write-Output sends the text as an object to the next pipeline command. If no later pipeline command is provided, the Write-Output also outputs the text to the screen." [source:here](https://morgantechspace.com/2021/02/write-output-vs-write-host-in-powershell.html) but I don't this this is the intention in these scripts?

fey101 (changes requested):
The linter is a great idea. Left a couple of comments.

scripts/TenantSetup/InitTenant.ps1
georgend: Added a new method `Write-ColorOutput` that correctly handles color output before calling `Write-Output`

scripts/TenantSetup/RaptorUtils.ps1
georgend: My thoughts were that suppressions should be close to the suppression point, and in this case the closest to the suppression point is at the top of the file.

fey101: Are there common suppression that cut across files? Should they be in a single file? For PSScriptAnalyzer, the file is called `PSScriptAnalyzerSettings.psd1` . I found a script that talks about the goal am hoping we can achieve here https://devblogs.microsoft.com/scripting/visual-studio-code-editing-features-for-powershell-development-part-2/#psscriptanalyzer-integration

scripts/applyReadPermissionsToManifest.ps1
```diff
@@ -109,4 +109,4 @@ $manifest |
   ConvertTo-Json -Depth 10 |
   Out-File -FilePath $outputPath
 
-Write-Host "Updated manifest file with $countDelegated delegated and $countApplication application permissions!" -ForegroundColor Green
\ No newline at end of file
+Write-Output "Updated manifest file with $countDelegated delegated and $countApplication application permissions!" -ForegroundColor Green
```
fey101, line 112: Did you forget to update this to use Write-ColorOutput

scripts/generateCSharpModels.ps1
```diff
@@ -1,19 +1,20 @@
...
-Write-Host "Path to typewriter tool: $typewriterPath"
+Write-Output "Path to typewriter tool: $typewriterPath"
 
 # download metadata
 $metadataDir = Join-Path $env:BUILD_SOURCESDIRECTORY "metadata"
 mkdir $metadataDir
 $metadataFile = Join-Path $metadataDir "metadata.xml"
 Invoke-WebRequest -Uri $env:MetadataURL -OutFile $metadataFile
-Write-Host "Downloaded $env:MetadataURL to $metadataFile" -ForegroundColor Green
+Write-Output "Downloaded $env:MetadataURL to $metadataFile" -ForegroundColor Green
```
fey101, line 17: Did you forget to update this to use Write-ColorOutput?

scripts/generateCSharpModels.ps1
fey101: Did you forget to update this to use Write-ColorOutput?

fey101 (commented):
Did you forget to update the lines that specify color to use Write-ColorOutput?

scripts/generateCSharpModels.ps1
georgend: Yes. I missed this one.

baywet (approved):

scripts/TenantSetup/InitTenant.ps1
```diff
@@ -53,6 +54,6 @@ $nestedScripts | ForEach-Object {
     $scriptName = $_.Name
     $scriptPath = $_.FullName
 
-    Write-Host "Executing $scriptName at $scriptPath" -ForegroundColor Green
+    Write-ColorOutput -ForegroundColor Green "Executing $scriptName at $scriptPath"
```
fey101, line 57: @georgend , wouldn't it be much easier to use the exclude rule in  PSScriptAnalyzerSettings.psd1 file as show in attached screenshot instead of writing a function to replace `Write-Host`? Not unless Write-Host has negative side effects that we would like to avoid by using write-output. Which doesn't seem to be the case 
image

zengin: Based on the conversation here, I think we can continue using `Write-Host` for this particular set of scripts: https://github.com/PowerShell/PSScriptAnalyzer/issues/1118

But if the color is not adding too much value, we can just switch to `Write-Output` too. So I am OK for both :)

fey101 (approved):

PR #687, opened 2021-12-02
Add data Script for Alert in Read Default Tenant Data Script.

georgend:
Alerts are system generated and thus just read available alerts from tenant.
Noted issue with Delegated Permissions not being returned from DevX API.
File updated on Azure Blob Storage.


TestsCommon/CSharpKnownIssues.cs
```diff
@@ -642,8 +641,6 @@ public static class CSharpKnownIssues
                 { "get-event-multiple-locations-csharp-V1-executes", MissingDataKnownIssue with { GitHubIssue = "https://github.com/microsoftgraph/msgraph-sdk-raptor/issues/610" } },
                 { "get-instances-csharp-V1-executes", MissingDataKnownIssue with { GitHubIssue = "https://github.com/microsoftgraph/msgraph-sdk-raptor/issues/610" } },
 
-                { "get-alert-csharp-V1-executes", MissingDataKnownIssue with { GitHubIssue = "https://github.com/microsoftgraph/msgraph-sdk-raptor/issues/611" } },
```
zengin, line 645: instead of removing this from the known issues list, please change the root cause to missing permission scopes.

zengin (changes requested):

TestsCommon/CSharpKnownIssues.cs
georgend: On further investigation, Permissions are returned: 
Seems DevX expects
`https://graphexplorerapi.azurewebsites.net/permissions?requesturl=/security/alerts&method=GET`
and not
`https://graphexplorerapi.azurewebsites.net/permissions?requesturl=security/alerts&method=GET`

The issue evolves into the data being dismissed/deleted eventually by the service. Microsoft Graph doesn't provide any assurances regarding how long alerts will stay stored.

zengin (approved):

PR #703, opened 2021-12-07
Section Group data scripts

georgend:
Add Data Scripts for Section Group Data. 
Add correct identifiers to identifiers file. 
Remove fixed issues from Known Issues.


zengin:
@georgend thanks for allowing me to iterate on your PR for the PR checks. GitHub doesn't kick off subsequent workflows even with a PAT. I raised a ticket. Based on their response, I will retract all the experimental commits here.

cc: @fey101 FYI..

.github/workflows/create-reports.yml
```diff
@@ -40,21 +40,13 @@ jobs:
...
\ No newline at end of file
+      uses: zengin/commit@v3.0.2
+      with:
+        directory: ${{ env.BUILD_SOURCESDIRECTORY }}/msgraph-sdk-raptor
+        token: ${{ secrets.PAT }}
+        branch: ${{ env.CURRENT_BRANCH }}
+        message: 'Update reports'
+        force: 'true'
+        name: 'Mustafa Zengin'
+        email: 'muzengin@microsoft.com'
```
nikithauc, line 52: Is it okay that this email is public?

.github/workflows/create-reports.yml
```diff
@@ -40,21 +40,13 @@ jobs:
...
-        git remote add origin https://$env:PAT@github.com/microsoftgraph/msgraph-sdk-raptor.git
-
-        git push origin $env:CURRENT_BRANCH
-      shell: pwsh
-      working-directory: ${{ env.BUILD_SOURCESDIRECTORY }}/msgraph-sdk-raptor
-      env:
-        CURRENT_BRANCH: ${{ env.CURRENT_BRANCH }}
-        PAT: ${{ secrets.PAT }}
\ No newline at end of file
+      uses: zengin/commit@v3.0.2
```
nikithauc, line 44: What does this do?

zengin: It was addressing the two issues I opened here in a fork:

https://github.com/actions-x/commit/issues/2
https://github.com/actions-x/commit/issues/3

But @RikudouSage is so kind to take an action really quick to fix these, so I will update the reference once the PRs are closed.

.github/workflows/create-reports.yml
zengin: already public in the git history :)

zengin:
woot! Suggestion from GitHub support worked and UnitTests are run on a commit from GitHub Actions.

It looks like I need to remove "create-reports" from required checks though as it won't always be run.

@georgend could you please address the PS linting errors in this PR?

georgend:
@zengin issue fixed.

PR #707, opened 2021-12-08
Fix code rebase error

fey101:
The code rebase error, was such that the `get-externalgroup-csharp-V1-executes` was duplicated in the file, causing pipeline run issues. After noticing the failure in `create reports` run, and fixing the issue locally, the commit fixing this was not included in the merged code as auto-merge had already happened.
auto-merge was enabled and create reports though it failed is not a required check. We will probably need to retain it as a required check. 
cc @zengin


baywet (approved):

zengin:
@fey101, I think we can abandon this PR as it is a subset of #712

PR #709, opened 2021-12-08
Handle Edge case for delta timestamp.

georgend:
Swap out this date for to now, which is similar to latest.
For new tenants, we cannot pick an arbitrary date since we won't know the state of the tenant.


zengin (changes requested):
It looks like the original date is generated at the time of snippet generation given the recent date. You may want to have a regex replace instead of string replace.

georgend:
Nice catch. Letme change to that.

PR #785, opened 2022-01-21
add missing data: online meeting

### LINKED ISSUE #573: Missing Data: OnlineMeeting
- [ ] get-custom-questions-csharp-Beta-executes 
- [ ] get-registration-csharp-Beta-executes
- [ ] list-registratrants-csharp-Beta-executes

zengin:
- fixes #573 
- Permission description doesn't exist for OnlineMeetingArtifact.Read.All, the new root cause for `get-attendee-report-csharp-V1-executes` becomes https://github.com/microsoftgraph/microsoft-graph-devx-content/issues/178


scripts/TenantSetup/TenantHttpData/onlineMeeting/onlineMeeting.json
```diff
@@ -0,0 +1,5 @@
+{
+    "startDateTime":"2030-07-12T14:30:34.2444915-07:00",
```
fey101, line 2: Any reason why this is not relative date? I will approve this but hopefully we can get it to relative date to future-proof the tests.

fey101 (approved):

scripts/TenantSetup/TenantHttpData/onlineMeeting/onlineMeeting.json
zengin: Why does it have to be relative? We only need the meeting to exist for its ID.

fey101: Say the script is running in 2031 in an automated fashion without monitoring? Wouldn't we need the start date to make sense such that the creation of this datapoint remains automatic without intervention? If the script was to run in 2031, the meeting might fail to create with error startdate must be later than current date, right? If this was a log, maybe it wouldn't matter, but if such validations are made by the api, then this will fail in the future.

zengin: I think I wasn't optimistic enough about this script :). My assumption was that 8 years was a long time for this script to stay alive, feel free to send a PR to make that longer.

zengin: That said, Outlook allows creating meetings in the past as well, though it shows a client side message like this:

image

fey101: As long as it allows it, then this should be okay. Thanks for clarifying.

PR #799, opened 2022-01-26
Move PowerShell tests to an execution matrix.

georgend:
Simplify PowerShell tests to use Azure Pipeline Strategies. 
The tests only differ by the name of the csproj to execute. 
This allows us to monitor the tests in the same run panel and to run them in parallel, reducing execution time when build agents are available. 
Also move the tests to smaller less resource intensive build agents to avoid Azure resource exhaustion limits. 

High Level Tests View
image

Detailed Test View 
image

***Misc Fixes***
Make test libraries non-CLSCompliant to remove build warning. 
Mark all tests as parallelizable.


TypeScriptBetaTests/AssemblyInfo.cs
```diff
@@ -0,0 +1,4 @@
+﻿using System;
+using NUnit.Framework;
+[assembly: Parallelizable(ParallelScope.All)]
```
baywet, line 3: this might be an issue since we're actually writing files to disk. CC @rkodev

georgend: I can remove it for typescript just to be safe.

georgend: Removed the offending attribute.

baywet (approved):

TypeScriptBetaTests/AssemblyInfo.cs
zengin: It is probably going to be OK to parallelize because TypeScript will be compiled as one unit at the beginning, and the tests will only report the outcome of the run. That said, keeping PRs scoped to intended change (Powershell tests in this case) is always a good idea ;)

PR #822, opened 2022-02-08
Remove beta compilation fixed issues

fey101:


TestsCommon/CSharpKnownIssues.cs
```diff
@@ -92,14 +74,9 @@ public static class CSharpKnownIssues
                 { "post-chatmessage-2-csharp-V1-compiles", IdentitySetAndIdentityShouldNestAdditionalDataKnownIssue},
 
                 { "create-educationrubric-from-educationassignment-csharp-V1-compiles", EducationAssignmentRubricContainsTargetPreprocessorKnownIssue},
-                { "create-directorysetting-from-settings-for-guests-csharp-Beta-compiles", new KnownIssue(Category.Metadata, "GroupSetting undefined in metadata")},
                 { "delete-educationrubric-from-educationassignment-csharp-V1-compiles", EducationAssignmentRubricContainsTargetPreprocessorKnownIssue},
                 { "add-educationcategory-to-educationassignment-csharp-V1-compiles", EducationAssignmentRubricContainsTargetPreprocessorKnownIssue},
 
-                { $"event-accept-csharp-{version}-compiles", EventActionsShouldNotBeReorderedKnownIssue },
-                { $"event-decline-csharp-{version}-compiles", EventActionsShouldNotBeReorderedKnownIssue },
-                { $"event-tentativelyaccept-csharp-{version}-compiles", EventActionsShouldNotBeReorderedKnownIssue },
```
zengin, line 101: We need to understand how these issues are fixed. It shouldn't be passing until kiota based SDKs are released.

zengin (changes requested):

TestsCommon/CSharpKnownIssues.cs
fey101: The kiota preview is not yet out as confirmed from Andrew so I've removed these testcases as the passing appears ephemeral just as you had suspected.

zengin (approved):

Issue #824, opened 2022-02-08
Merge msgraph-sdk-raptor-compiler-lib into TestsCommon


zengin:
Both of these projects serve more or less the same purpose and we don't need to maintain the indirections between them.

Issue #201, opened 2021-03-02
[Java] test performance improvements


baywet:
Right now the java raptor tests take upwards of an hour for beta. This trips the ADO time limit and makes the tests flaky on top of making it painful to use.
This is because we currently create a shell project, copy the snippet, and fire the compilation for every single test. The gradle process start and reporting takes a lot of time, and is repeated for each snippet.

If we instead had a gradle multi-project structure, with one shell (deps and common configuration) and one subproject per snippet, we'd be able to fire the compilation once. Saving a lot of cold start and stop time per unit test.

That pre-compilation sequence would be fired in the test initialization method (once for the data set) and the unit test would only match the result (stored in a file?) and report it up to Nunit. 
[AB#8389](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/8389)

zengin:
cleaning up open issues: no longer an issue after #825

PR #825, opened 2022-02-09
Java improvements

zengin:
This PR is a side product of exploring different kinds of compilation patterns for different languages in Raptor.

After observing tremendous improvement in the TypeScript compilation when all files were combined in a single `tsc` run (over an hour vs couple of minutes), I wanted to give that a try for Java to see if we can get a similar result. I found that java compiler bails out early if multiple files are attempted to be compiled in a single compile attempt. I searched for a compiler parameter to work around it, but couldn't find one. While investigating this, I have realized that calling `javac` directly with a flat directory hierarchy improves performance over `gradle` builds. We don't need the retries anymore and it is significantly faster both on Windows and Linux. Down that rabbit hole, I also gave ramdisks a try as well, that didn't improve the total run time that much, so I reverted that change back for clarity.

These changes allowed us to reclaim the paid Azure hardware for pipelines, we are now able to run all Java pipelines on `ubuntu-latest` once again. That said, there is still significant opportunities for improvement, so I will keep the issue #201 on performance open.

One avenue we can explore is:
Interestingly, `javac` doesn't bail out early if everything is dumped into a single file and compiles with ~8000 errors within seconds, we can potentially parse the errors back to understand which failures belong to which snippet, but there is a risk that one snippet could affect another (e.g. an unclosed string literal overflowing to the next snippet). Different files give us the isolation, but I couldn't find a way to force `javac` to compile all files in a single run. I get the same behavior in IntelliJ, so I'm afraid there may not be a way. I have this SO post open, yet to be answered: https://stackoverflow.com/questions/70778646/how-to-prevent-gradle-build-from-failing-early

##### One time setup:
- A project directory is created
- gradle is set up in that directory

##### Test case:
- Each snippet is transformed and dumped into an App.java file
- App.java file is compiled using `gradle build` sequentially
- Compiler error is mapped to CSharp's `Diagnostic` object

#### One time setup:
- A project directory is created
- gradle is set up to download the dependecies from maven central
- `gradle download` is called to dump all `jar` files of the transitive dependencies into the `lib` directory
- [Preview case] locally built core and service libraries are copied into the `lib` directory
- Each snippet is transformed and dumped into their own file

#### Test Case:
- For each snippet `javac` is called with dependecies in `lib` [parallelized]
- `javac` output is directly served without transformation into `Diagnostic` object

This is also a partial #824 as we are working towards consolidating `msgraph-sdk-raptor-compiler-lib` project into `TestsCommon` with a focus on understanding different compilation/execution requirements for each language that onboards into Raptor.

### Show me the data:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/muzengin/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/muzengin/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

  | Before (minutes) | with javac (minutes) | javac (speedup) | with RamDisk (minutes) | with RamDisk (speedup)
-- | -- | -- | -- | -- | --
V1 Stable Test Run Time | 78 | 26 | 3.00 | 29 | 2.69
V1 Stable Complete Pipeline | 80 | 27.5 | 2.91 | 30 | 2.67
V1 Known Issues Test Run Time | 4.2 | 2.7 | 1.56 | 2.7 | 1.56
V1 Known Issues Complete Pipeline | 6 | 4.5 | 1.33 | 4 | 1.50
Preview Stable Test Run Time | 78 | 25.5 | 3.06 | 25.5 | 3.06
Preview Stable Complete Pipeline | 80 | 29.5 | 2.71 | 29.5 | 2.71
Preview Known Issues Test Run Time | 4 | 2.5 | 1.60 | 2.5 | 1.60
Preview Known Issues Complete Pipeline | 9 | 6.5 | 1.38 | 6.5 | 1.38
Beta Stable Test Run Time | 128 | 41.5 | 3.08 | 54 | 2.37
Beta Stable Complete Pipeline | 130 | 43 | 3.02 | 52 | 2.50
Beta Known Issues Test Run Time | 10 | 7.5 | 1.33 | 7.5 | 1.33
Beta Known Issues Complete Pipeline | 12 | 9 | 1.33 | 9 | 1.33

</body>

</html>


TestsCommon/CompilationOutputMessage.cs
```diff
@@ -4,21 +4,21 @@ namespace TestsCommon;
...
 /// </summary>
-/// <param name="Model">Compilation result</param>
+/// <param name="CompilationResult">Compilation result</param>
 /// <param name="Code">Code that was attempted to be compiled</param>
 /// <param name="DocsLink">documentation page where the snippet is shown</param>
 /// <param name="KnownIssueMessage">message for known issue</param>
 /// <param name="IsKnownIssue">whether issue is known</param>
 /// <param name="Language">compilation language</param>
-public record CompilationOutputMessage(CompilationResultsModel Model, string Code, string DocsLink, string KnownIssueMessage, bool IsKnownIssue, Languages Language = Languages.CSharp)
+public record CompilationOutputMessage(string CompilationResult, string Code, string DocsLink, string KnownIssueMessage, bool IsKnownIssue, Languages Language = Languages.CSharp)
```
zengin, line 13: CompilationResult is now a string that can be served directly from different types of compilers. @rkodev, I think we can remove the indirection to parse the results into `Diagnostic` and then back to string here.

azure-pipelines/java-v1-compile-tests.yml
```diff
@@ -31,25 +31,22 @@ resources:
...
+pool:
+  vmImage: 'ubuntu-latest'
+
 variables:
   buildConfiguration: 'Release'
 
-jobs:
-  - job: CompileJavaV1Snippets
-    displayName: 'V1 Java Snippet Compilation Tests'
-    pool: 'MsGraphBuildAgentsLinux'
```
zengin, line 40: build agents are reclaimed.

msgraph-sdk-raptor-compiler-lib/Models/CompilationResultsModel.cs
```diff
@@ -1,5 +1,26 @@
 ﻿namespace MsGraphSDKSnippetsCompiler.Models;
 
-public record CompilationResultsModel(bool IsSuccess, IEnumerable<Diagnostic> Diagnostics, string MarkdownFileName);
-#pragma warning disable CA1801 // Review unused parameters (false positive, seems to be fixed in vNext of dotnet: https://github.com/dotnet/roslyn-analyzers/pull/4499/files)
-public record ExecutionResultsModel(CompilationResultsModel CompilationResult, bool Success, string ExceptionMessage = null);
+public record CompilationResultsModel(bool IsSuccess, IEnumerable<Diagnostic> Diagnostics, string MarkdownFileName)
+{
+    public override string ToString()
```
zengin, line 5: string representation is moved to the model object.

msgraph-sdk-raptor-compiler-lib/Models/LanguageTestData.cs
```diff
@@ -17,18 +17,35 @@
...
+    string FileName,
+    string DllPath,
+    string JavaCoreVersion,
+    string JavaLibVersion,
+    string JavaPreviewLibPath,
+    string TestName,
+    string Owner,
+    string FileContent)
+    {
+        public string JavaClassName => string.Join("", FileName
```
zengin, line 36: We needed unique class names for each snippet since we now dump all files into a single directory at once.

TestsCommon/ProcessSpawner.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    // create a static method that spawns a process and waits for a timeout and returns stdout and stderr as a tuple
+    public static async Task<(string stdout, string stderr)> SpawnProcess(string command, string arguments, string workingDirectory, int timeout)
+    {
+        _ = command ?? throw new ArgumentNullException(nameof(command));
+        _ = arguments ?? throw new ArgumentNullException(nameof(arguments));
+        _ = workingDirectory ?? throw new ArgumentNullException(nameof(workingDirectory));
+
+        ProcessStartInfo startInfo = new ProcessStartInfo(command, arguments);
+        startInfo.UseShellExecute = false;
```
baywet, line 15: nit; object initialization instead

TestsCommon/JavaTestRunner.cs
```diff
@@ -1,12 +1,96 @@
...
+    toCopy 'com.microsoft.graph:microsoft-graph-beta:--libversion--'
+}
+
+task download(type: Copy) {
+    from configurations.toCopy
+    into 'lib'
+}
+";
+
+    private const string Deps = @"toCopy 'com.google.guava:guava:30.1.1-jre'
```
baywet, line 85:
```suggestion
    private const string Deps = @"toCopy 'com.google.guava:guava:31.0.1-jre'
```

TestsCommon/JavaTestRunner.cs
```diff
@@ -1,12 +1,96 @@
...
+    toCopy 'com.google.code.gson:gson:2.8.6'
```
baywet, line 86:
```suggestion
    toCopy 'com.google.code.gson:gson:2.8.9'
```

TestsCommon/JavaTestRunner.cs
```diff
@@ -1,12 +1,96 @@
...
+    toCopy 'com.squareup.okhttp3:okhttp:4.9.1'
```
baywet, line 87:
```suggestion
    toCopy 'com.squareup.okhttp3:okhttp:4.9.3'
```

TestsCommon/JavaTestRunner.cs
```diff
@@ -1,12 +1,96 @@
...
+    toCopy 'com.azure:azure-identity:1.2.5'";
```
baywet, line 88:
```suggestion
    toCopy 'com.azure:azure-identity:1.4.3'";
```

TestsCommon/JavaTestRunner.cs
```diff
@@ -62,17 +146,143 @@ public class App
...
+                RedirectStandardOutput = true,
+                RedirectStandardError = true,
+                CreateNoWindow = true,
+                WorkingDirectory = compilationDirectory
+            }
+        };
+
+        gradleProcess.Start();
+        int dependencyDownloadTimeoutInSeconds = 120;
+        var hasExited = gradleProcess.WaitForExit(dependencyDownloadTimeoutInSeconds * 1000);
```
baywet, line 250: I know the wait for exit async method doesn't accept a timeout. But should we consider running a timer and cancel the task instead? or running two tasks in parallels (task.delay + task.any pattern)?
Doing this could improve the degree of concurrency, but if I understand things right, this method is only run once per test run instead of once per test case now, correct?

baywet (approved):
This is amazing work! Thanks for taking the time to fix it!🙌🥳
This approach is going to benefit every language we'll be adding in the future (Go next? 😉)

I did have this improvement to try in my Todo list since forever but never actually got time to get to it. 

Just a few minor comments/questions on my side.

TestsCommon/JavaTestRunner.cs
zengin: @baywet what do you think if we create a dummy java project and read dependencies from there so that dependabot handles these upgrades for us?

baywet: that'd be even better! but I didn't want to mix concerns too much on this PR.

zengin: cool, created an issue for this: https://github.com/microsoftgraph/msgraph-sdk-raptor/issues/826

TestsCommon/JavaTestRunner.cs
zengin: Thanks for the suggestion, I haven't changed this process as this runs only once per test suite. But I applied the suggestion into the process spawner which runs per test case.

baywet: Out of curiosity did you notice any further improvements?

baywet (approved):

TestsCommon/JavaTestRunner.cs
zengin: I observed 5-10% gain in long-running tests 🎉 That said, I observed that a few tests started hitting the timeout in the local run, this wasn't happening in the synchronous runs. I believe it is due to increased concurrency. I will watch the tests over the next week and see if this causes any flakiness in the test results.

PR #832, opened 2022-02-10
gets java build files from dependabot-managed directories

### LINKED ISSUE #826: Create a dummy Java project for automated dependabot upgrades
@baywet what do you think if we create a dummy java project and read dependencies from there so that dependabot handles these upgrades for us?

_Originally posted by @zengin in https://github.com/microsoftgraph/msgraph-sdk-raptor/pull/825#discussion_r802886887_

zengin:
- fixes #826
- removes inline build.gradle file contents
- adds one build.gradle for each type of run: v1, beta, preview
- removes scripts that gets the latest published package (instead, relies on dependabot's management)
- keeping versions old to get a faster feedback from dependabot config


baywet (approved):
what a cleanup! thanks for taking the time to do that.

PR #853, opened 2022-02-17
Fix pipelines after Typewriter update

andrueastman:
This PR fixes the metadata transform task as its now broken after the update made via https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/676 and https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/672

### DIFF
--- a/scripts/generateCSharpModels.ps1
+++ b/scripts/generateCSharpModels.ps1
@@ -9,7 +9,7 @@ $raptorUtils = Join-Path $PSScriptRoot "./TenantSetup/RaptorUtils.ps1" -Resolve
 $outputPath = Join-Path -Path $env:BUILD_SOURCESDIRECTORY -ChildPath "output"
 Write-Output "Path to typewriter output $outputPath"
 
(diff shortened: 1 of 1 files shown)


scripts/generateCSharpModels.ps1
```diff
@@ -9,7 +9,7 @@ $raptorUtils = Join-Path $PSScriptRoot "./TenantSetup/RaptorUtils.ps1" -Resolve
 $outputPath = Join-Path -Path $env:BUILD_SOURCESDIRECTORY -ChildPath "output"
 Write-Output "Path to typewriter output $outputPath"
 
-$typewriterDirectory = Join-Path -Path $env:BUILD_SOURCESDIRECTORY -ChildPath "MSGraph-SDK-Code-Generator" "src" "Typewriter" "bin" $env:BuildConfiguration "net5.0"
```
zengin, line 12: we should probably make this version-agnostic, but Typewriter is going away soon, so maybe not worth spending time doing it.

zengin (approved):

scripts/generateCSharpModels.ps1
andrueastman: True. This is something to think about for the next version.

baywet:
Thanks for fixing that. I forgot that Typewriter was used in raptor

PR #854, opened 2022-02-17
Modify validation for TestType in Class RunSettings

fey101:
Modify validation for TestType in Class RunSettings so as to accept default testContext.Parameters values without throwing an exception

Original behaviour:
Running JavaV1Tests locally revealed the bug below. The runsettings in associated file is defined as below:
```C#
private static RunSettings runSettings => new RunSettings(TestContext.Parameters)
        {
            Version = Versions.V1,
            Language = Languages.Java,
            TestType = TestType.CompilationStable
        };
```
This initially? passes the default values from Test.runsettings whereby the testype is defined as `<Parameter name="TestType" value="---TestTypePlaceholder---" />` which is caught in the else block [here](https://github.com/microsoftgraph/msgraph-sdk-raptor/blob/dev/TestsCommon/RunSettings.cs#L86) and throws an exception $"Unexpected test type specified: {testType}"

Currently:
Added code to validate that runsettings are provided via the runsettings file or via an override. Override settings is optional but when the file contains placeholder values with dashdash, then override settings must be provided, Only override settings when the file either contains placeholder values with dashdash or when the parameter values in the file are not provided. Otherwise use values provided in the file.


TestsCommon/RunSettings.cs
```diff
@@ -103,4 +117,19 @@ private static string InitializeParameter(TestParameters parameters, string para
             return null;
         }
     }
+    protected static bool isPlaceholderValueOrEmpty(string propertyName){
+        return string.IsNullOrEmpty(propertyName) || propertyName.Contains(DashDash);
+    }
+
+    protected static T replaceUnprovidedSettingsValueWithOverride<T>(RunSettings overrideSettings, string propertyName){
```
zengin, line 124: nit: PascalCase and private
```suggestion
    private static T ReplaceUnprovidedSettingsValueWithOverride<T>(RunSettings overrideSettings, string propertyName){
```

TestsCommon/RunSettings.cs
```diff
@@ -103,4 +117,19 @@ private static string InitializeParameter(TestParameters parameters, string para
             return null;
         }
     }
+    protected static bool isPlaceholderValueOrEmpty(string propertyName){
```
zengin, line 120: nit: PascalCase and private
```suggestion
    private static bool IsPlaceholderValueOrEmpty(string propertyName){
```

zengin (changes requested):
Could you please kick off a run for staging pipelines on this branch? I want to make sure that the changes around `DllPath` doesn't break anything there.

fey101 (commented):
I have run the staging-end-to-end pipelines for both beta and V1 and they run to completion. I had also tested this locally and the good thing with our way of initializing the override settings is that all variables are first initialized to null so that even though the override settings don't explicitly include a dll path, the override settings have the property DllPath initialized to null. I thought I should keep it that way since we are guaranteed those four critical properties of RunSettings file will always be intentional. And as long as it's set on the file to a value other than null or the placeholder value, it won't be overridden. Sample pipeline here:https://microsoftgraph.visualstudio.com/Graph%20Developer%20Experiences/_build/results?buildId=68529&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=04da9304-4993-52d9-c871-c21bca046544

TestsCommon/RunSettings.cs
fey101: Wouldn't Protected be better so that the methods can be inherited by child classes?

TestsCommon/RunSettings.cs
fey101: I preferred protected over private to enable inheritance of the methods in child classes.

TestsCommon/RunSettings.cs
zengin: That change can be made when we decide to use it with inheritance, which is not the case at the moment. Keeping it private is self-documenting to understand in which contexts this piece of code is called.

zengin (approved):

PR #860, opened 2022-02-18

zengin:
This is an automated PR.


fey101 (approved):
No change at all in the package versioning. Merging it in since it would just be more complex to try and do otherwise. On the other hand there's no particular harm in merging it.

PR #901, opened 2022-03-15
Bug fix on identifiers replacement for powershell

timayabi2020:
Powershell snippets that have more than one placeholder that need to be replaced are failing.
In this scenario only one place holder will be replaced with the corresponding identifier in the identifiers.json file.
To correct this, a recursive call of the function that does those replacements has been implemented


TestsCommon/Models/IdentifierReplacer.cs
```diff
@@ -218,6 +228,50 @@ private string ReplaceIdsFromIdentifiersFile(string input)
 
         return input;
     }
+    /// <summary>
+    /// Replaces ID placeholders of the form {nameId} by looking up name in the IDTree.
+    /// Recursively call the function based on the number of Ids that needs to be replaced on the snippet
+    /// </summary>
+    /// <param name="input"></param>
```
fey101, line 235: This feels quite similar to the existent `    private string ReplaceIdsFromIdentifiersFile(string input)
` function. Is there a difference introduced? Could refactoring help in putting what is common together and what is not separately for clarity?

timayabi2020: Yes there is a difference on lines 257 (Throw exception only during replacement process of the last placeholder in the snippet) and 269 (Recursive bit). However, I will still look at further refactoring of the almost similar previous function that was being used

PR #905, opened 2022-03-17
Power shell identifiers replacement bug fix

timayabi2020:
Powershell snippets that have more than one placeholder that need to be replaced are failing.
In this scenario only one place holder will be replaced with the corresponding identifier in the identifiers.json file.
To correct this, a recursive call of the function that does those replacements has been implemented.


zengin (changes requested):
please add a unit test with an example. I don't fully understand what the bug is and how this PR resolves the bug.

timayabi2020:
> please add a unit test with an example. I don't fully understand what the bug is and how this PR resolves the bug.

Unit test has been added as per request

UnitTests/PowershellIdentifierReplacerTest.cs
```diff
@@ -0,0 +1,72 @@
...
+    public void SnippetWithMultiplePlaceHoldersShouldThrowAnException(string testSnippet, string expectedTestSnippet) => Assert.Throws<InvalidDataException>(()=>_idReplacer.ReplaceIds(testSnippet, Languages.CSharp.ToString()));
+
+    /// <summary>
+    ///  Snippet with a single placeholder should pass (indicating expected and recieved results are equal) as long as the language is not identified as powershell
+    /// </summary>
+    /// <param name="snippetWithSinglePlaceHolders"></param>
+    /// <param name="expectedSnippetWithSinglePlaceHolders"></param>
+
+    [TestCase(snippetWithSinglePlaceHolders, expectedSnippetWithSinglePlaceHolders)]
+    public void SinglePlaceHolderSnippetIdentifiedAsPowerShellShouldPass(string snippetWithSinglePlaceHolders, string expectedSnippetWithSinglePlaceHolders) => Assert.AreEqual(_idReplacer.ReplaceIds(snippetWithSinglePlaceHolders, Languages.PowerShell.ToString()), expectedSnippetWithSinglePlaceHolders);
```
georgend, line 58: Using this method style affects readability:-
- line is too long across the screen. 
- Breaking it down into Arrange step, Act Step and Assert step, instead of a one liner would enhance readability and comprehension.

georgend (changes requested):

TestsCommon/Models/IdentifierReplacer.cs
```diff
@@ -207,13 +211,45 @@ private string ReplaceIdsFromIdentifiersFile(string input)
             if (localTree == null
                 || localTree.Value.EndsWith($"{idType}>", StringComparison.Ordinal)) // placeholder value, e.g. <teamsApp_teamsAppDefinition> for teamsApp->teamsAppDefinition
             {
-                throw new InvalidDataException($"no data found for id: {id} in identifiers.json file");
+                string errorMessage = $"no data found for id: {id} in identifiers.json file";
+                if (language == Languages.PowerShell.ToString())
```
georgend, line 215: To avoid boxing, remove the `ToString` and change language to be of enum type `Languages`.

TestsCommon/Models/IdentifierReplacer.cs
```diff
@@ -186,17 +185,22 @@ internal static string RegexReplaceEdgeCases(List<(Regex SearchRegex, string Rep
...
     ///   {site-id} is replaced by root->site entry from the IDTree.
     ///   {list-id} is replaced by root->site->list entry from the IDTree.
+    ///   For powershell snippets with more than one placeholder, a recursive call is done
+    ///   to avoid a scenario where only one placeholder is replaced and an exception is thrown
+    ///   while replacing the subsequent placeholders
     /// </summary>
     /// <param name="input">String containing ID placeholders</param>
     /// <returns>input string, but its placeholders replaced from IDTree</returns>
-    private string ReplaceIdsFromIdentifiersFile(string input)
+    private string ReplaceIdsFromIdentifiersFile(string input,string language)
```
georgend, line 194: Change `language` to be of type `Languages`

TestsCommon/CSharpTestRunner.cs
```diff
@@ -178,7 +178,7 @@ private static string GetResultVariable(string codeToCompile)
     /// <returns>code to be executed</returns>
     private static (string, string) GetCodeToExecute(string fileContent)
     {
-        var (codeToCompile, codeSnippetFormatted) = GetCodeToCompile(IdentifierReplacer.Instance.ReplaceIds(fileContent));
+        var (codeToCompile, codeSnippetFormatted) = GetCodeToCompile(IdentifierReplacer.Instance.ReplaceIds(fileContent,Languages.CSharp.ToString()));
```
georgend, line 181: Pass in Languages without turning to a string.

georgend (changes requested):
Minor changes required

TestsCommon/Models/IdentifierReplacer.cs
timayabi2020: This is also done. The type has been changed from language to Languages

TestsCommon/Models/IdentifierReplacer.cs
```diff
@@ -207,13 +211,45 @@ private string ReplaceIdsFromIdentifiersFile(string input)
...
+
             }
             else
             {
                 currentIdNode = localTree;
                 input = ReplaceFirst(input, id, currentIdNode.Value);
             }
+            count++;
+        }
+        if (language == Languages.PowerShell.ToString())
```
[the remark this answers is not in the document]
timayabi2020: This is also done

TestsCommon/Models/IdentifierReplacer.cs
timayabi2020: This is also done

TestsCommon/Models/IdentifierReplacer.cs
```diff
@@ -186,17 +185,22 @@ internal static string RegexReplaceEdgeCases(List<(Regex SearchRegex, string Rep
...
+    private string ReplaceIdsFromIdentifiersFile(string input,string language)
     {
         if (input == null)
         {
             throw new ArgumentNullException(nameof(input));
         }
 
+        int count=0;
         var matches = _identifierRegex.Matches(input);
+        List<Match>matchList = matches.ToList();
```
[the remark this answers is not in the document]
timayabi2020: Removed ToList and made use of Count

TestsCommon/CSharpTestRunner.cs
timayabi2020: Also done

TestsCommon/Models/IdentifierReplacer.cs
```diff
@@ -55,10 +55,10 @@ internal static readonly List<(Regex SearchRegex, string ReplacementContent)> Ed
         )
     };
 
-    public IdentifierReplacer(IDTree tree)
+    public IdentifierReplacer(IDTree tree,string language)
```
georgend, line 58: Switch to the language Enum here as well.

TestsCommon/Models/IdentifierReplacer.cs
```diff
@@ -55,10 +55,10 @@ internal static readonly List<(Regex SearchRegex, string ReplacementContent)> Ed
...
     {
         this._tree = tree;
-        _identifierRegex = IdentifierRegexes.CsharpIdRegex;
+        _identifierRegex =  language==Languages.CSharp.ToString()?IdentifierRegexes.CsharpIdRegex:IdentifierRegexes.PowerShellIdRegex;
```
georgend, line 61: Switch to the language enum here as well eliminate the `ToString()`

zengin:
@timayabi2020 thanks for writing the unit tests.

There is no need to have a code change for multiple ID scenarios, because the same problem exists in C# implementation and we already resolved it by iterating on the tree and regex matches at the same time. We should not call the replaceId function recursively as the iteration already covers that scenario.

I was able to revert your code changes and added user->event into the identifiers.json file and the test case `SnippetWithMultiplePlaceHoldersShouldPass` still passes.

timayabi2020:
> @timayabi2020 thanks for writing the unit tests.
> 
> There is no need to have a code change for multiple ID scenarios, because the same problem exists in C# implementation and we already resolved it by iterating on the tree and regex matches at the same time. We should not call the replaceId function recursively as the iteration already covers that scenario.
> 
> I was able to revert your code changes and added user->event into the identifiers.json file and the test case `SnippetWithMultiplePlaceHoldersShouldPass` still passes.

I had also explored that option locally but my question is, do we have the flexibility of doing such modifications on the identifiers.json file in cloud? There is a possibility that powershell snippets might  have more than two or three Id placeholders and I think it will be cumbersome to always manually add an object with its values under one primary object . e.g user->event->calender->etc->etc->etc.

zengin:
@timayabi2020 we have all the flexibility to change identifiers.json. The only thing that I'd suggest we watch out for is that the same hierarchy exists between C# and PowerShell. If that is not the case, we can reevaluate if we need a separate identifiers.json file for PowerShell tests. But given the high pass rate, I really doubt there is a huge difference between the two languages, if there is any.

timayabi2020:
> > @timayabi2020 we have all the flexibility to change identifiers.json. The only thing that I'd suggest we watch out for is that the same hierarchy exists between C# and PowerShell. If that is not the case, we can reevaluate if we need a separate identifiers.json file for PowerShell tests. But given the high pass rate, I really doubt there is a huge difference between the two languages, if there is any.
> 
> @zengin Since we have the flexibility to update the identifiers.json file, I have made an update to this PR to only include unit tests for powershell identifier replacement and updates on powershell known issues file. The other changes on Identifier replacement function has been rolled back too.

zengin (approved):

fey101 (approved):

georgend (approved):

PR #955, opened 2022-04-11
Execute CredScan on every PR

### LINKED ISSUE #864: Add CredScan to pipelines
This is to prevent any leaks of secrets in the source code.

georgend:
Fixes #864 

- Adds a CredScan task immediately after checkout
- Only scan raptor folders. 

### Other links
- https://secdevtools.azurewebsites.net/helpcredscan.html


MIchaelMainer:
No checking in Guardian files to public repos. They contain information about found vulnerabilities.

zengin:
@georgend can't we hook the CredScan task directly instead of running it as a CLI tool? I guess that would address the concern that @MIchaelMainer has raised.

MIchaelMainer:
And nice find on Guardian. There will be all sorts of good stuff in that as time goes on.

PR #973, opened 2022-04-13
Powershell knownissues update 130422

timayabi2020:
Updated the known issues file list.


scripts/TenantSetup/DefaultData/readDefaultData.ps1
```diff
@@ -265,7 +265,7 @@ $userEvent = Invoke-RequestHelper -Uri "users/$($identifiers.user._value)/calend
     Where-Object { $_.recurrence -ne $null } | # need a recurring event for "get instances" API call
     Select-Object -First 1
 $identifiers = Add-Identifier $identifiers @("event") $userEvent.id
-$identifiers = Add-Identifier $identifiers @("user", "event") $userEvent.id
+$identifiers = Add-Identifier $identifiers @("user", "event","outlookCategory","notebook","todoTaskList","onenoteSection","sectionGroup") $userEvent.id
```
zengin, line 268: I don't understand the intent with this change. Could you describe the issue this is resolving?

zengin (changes requested):

scripts/TenantSetup/DefaultData/readDefaultData.ps1
timayabi2020: @zengin I was attempting to add child nodes under user on the "identifiers.json" using that script but i've rolled back since one can add it manually.

zengin (approved):

PR #1055, opened 2022-05-19
Remove dependency on main method at compilation stage and reactivate

fey101:
parallelization

Closes #1003


TestsCommon/GoTestRunner.cs
```diff
@@ -149,12 +143,14 @@ public static async Task DumpGoFiles(IEnumerable<LanguageTestData> languageTestD
         foreach(var testData in languageTestData)
         {
             var (codeToCompile, _) = GetCodeToCompile(testData.FileContent);
-            var testNameAsFunctionName = ReplaceHyphensWithUnderscores(testData.TestName);
+            var testNameAsFunctionName = CultureInfo.CurrentCulture.TextInfo.ToTitleCase(
+                ReplaceHyphensWithUnderscores(testData.TestName).ToLower()
```
baywet, line 147:
```suggestion
                ReplaceHyphensWithUnderscores(testData.TestName).ToLowerInvariant()
```

baywet (changes requested):

zengin:
@fey101, this is taking way too long to run even in parallel mode on my local box, Also I can't get it to work if I don't run a single test in isolation with a manual `go build <snippet>.go` which I believe warms up the local cache for the SDK dependencies. I think we need to have a warm up step, as well as consider running all snippets in a single `go build` call and then parsing results (similar to what @rkodev did in TypeScript). `go build` bails out early with too many errors, so you may need to reuse @rkodev's hack for isolating passing tests.

zengin:
@fey101 with the cache warm up, the process took about an hour to run on Windows. With half an hour runtime on Linux, I think that this is acceptable for now. We can have another shot at performance at a later date.

Let's continue with building the pipelines and measure how long it takes to run the tests with a `ubuntu-latest` image.

zengin (approved):

zengin:
numbers...

on WSL:
```
Test Run Failed.
Total tests: 1141
     Passed: 116
     Failed: 1025
 Total time: 9.1118 Minutes
```

zengin:
@baywet, do you mind approving this PR if you don't have any other concerns? Since your last review, we added a cache warm up step which basically attempts to compile the first file in the folder, which creates a build cache and speeds up the subsequent test runs.

baywet:
Thanks for making the updates. In terms of build performance, I think you should have a look at this thread. I'm going to engage with the go core community but it's possible we uncover issues with the go compiler. https://github.com/microsoftgraph/msgraph-sdk-go/issues/129

PR #1073, opened 2022-05-26
known issues cleanup 5/25/2022

zengin:
partial #196 

- Triaged compilation issues for C#


zengin:
@timayabi2020 could you please check the "create reports" failure here? It looks related to your recent addition of PS issues to the reports.

zengin:
@timayabi2020 @fey101 ReportGenerator runs a check on known issues to make sure that there is a reference to the corresponding snippet in the documentation. It fails on known issues that don't actually exist in the documentation. This could be a legitimate scenario where a snippet is removed from the docs intentionally. We don't want to keep a stale reference in that case.

That said, PowerShell snippet generation is not stable, that is, it sometimes generates some of the snippets, and fails to do so in another week's run. We don't want to lose the work we have done previously in identifying the root causes for the known issues in this case.

So I introduced another list of languages to determine whether we should be running the check that the issues continued to be referenced. C# snippet generation is stable enough to handle that. PowerShell, not so much. So only C# is in that list.

PR #1078, opened 2022-05-27
Update Go Known Issues List

### LINKED ISSUE #1004: Print Unexpected Test Results to Quickly List Known Issues

fey101:
Closes #1004
- Updates the go known issues list as per a local recent run 
- Also Removes powershell known issues that no longer exist on the docs as highlighted by previously failing create reports task.


timayabi2020 (approved):

zengin:
We shouldn't have removed the PowerShell known issues that don't exist in the docs, because snippet generation for PowerShell is not stable. This will require putting them back once they are regenerated.

PR #1085, opened 2022-05-31
Use GitHub runner

MIchaelMainer:


.github/workflows/codeql-analysis.yml
```diff
@@ -12,7 +12,7 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, 1ES.Pool=1es-windows-github-latest]
```
zengin, line 15: any specific reason to switch from ubuntu to windows?

zengin: and how is the connection established between GitHub and Azure? GitHub would need a way to discover where to look at for the 1ES agents, I presume..

MIchaelMainer: > any specific reason to switch from ubuntu to windows?

No reason. I just made the Windows runner first to validate that it [ran successfully](https://github.com/microsoftgraph/msgraph-sdk-raptor/runs/6693445664?check_suite_focus=true). It did. I've since created an Ubuntu runner. I will update the PR.

The connection is managed by a GitHub app: 1ES Resource Management. It provides the github runner group to GitHub Actions.

zengin (approved):

baywet (approved):

PR #1089, opened 2022-06-02
Move failing tests to backup folder

### LINKED ISSUE #1084: Run Folder Level Compilation to Speed up Runs
- Runs taking over 1 hour to run with current implementation
- Goal is to optimize the runs for faster compilation

Requirements:
- [x] run go build on directory level instead of on individual file
- [x]  Move failed files onto backup directory
- [x]  Repeat steps above until the entire set of tests is completed( A single run will display errors from 1st 10 files)
- [ ]  Retry compiling on backup (optional)
- [x]  Report failed tests

fey101:
Closes #1084


TestsCommon/GoTestRunner.cs
```diff
@@ -171,38 +182,111 @@ public static async Task DownloadRequiredDependencies()
...
 
-    private static async Task CompileSnippet(LanguageTestData testData)
+    private async static Task<Dictionary<string, Collection<Dictionary<string, string>>>> CompileSnippet (IEnumerable<TestCaseData> testData)
     {
-        //Compile file
-        var compilePath = Path.Combine(CompilationDirectory, "src", testData.TestName + ".go");
-        var (stdout, stderr) = await ProcessSpawner.SpawnProcess(
+        //Compile folder
+        var compilationOutput = new Dictionary<string, Collection<Dictionary<string, string>>>();
+        int maxRetryCounts = 160;
```
rkodev, line 201: Given that its possible for the number of tests to fluctuate, we could add a dynamic retry ceiling i.e `number of files / 10`

TestsCommon/GoTestRunner.cs
```diff
@@ -171,38 +182,111 @@ public static async Task DownloadRequiredDependencies()
...
+                payload.Add(
+                    new Dictionary<string, string> {
+                        {"errorCode", errorCode},
+                        {"errorMessage", message}
+                    }
+                );
+
+                resultDict[filename] = payload;
+            }
+            return resultDict;
```
rkodev, line 254: minor: optional: reduce code complexity. exit on first failed if and avoid wrapping the second code block in an else

```c#
        if (!errors.Any())
            return resultDict;
        
        foreach (var errorline in errors)
        {
            if (!errorline.Contains("compiles.go")) continue;
            var items = errorline.Split(": ");
            var filename = items[0].Split(":").First().Split("src\\")[1];
            var message = errorline;
            var errorCode = items[1];

            var payload =  (resultDict.ContainsKey(filename)) ? resultDict[filename] :  new Collection<Dictionary<string, string>>();

            payload.Add(
                new Dictionary<string, string> {
                    {"errorCode", errorCode},
                    {"errorMessage", message}
                }
            );

            resultDict[filename] = payload;
        }
        return resultDict;
```

TestsCommon/GoTestRunner.cs
```diff
@@ -171,38 +182,111 @@ public static async Task DownloadRequiredDependencies()
...
         }
-        else
-        {
-            Assert.Pass();
+        else {
+            foreach (var errorline in errors)
+            {
+                if (!errorline.Contains("compiles.go")) continue;
+                var items = errorline.Split(": ");
+                var filename = items[0].Split(":").First().Split("src\\")[1];
```
rkodev, line 239: possible bug: directly accessing array index without checking length / size.

rkodev (changes requested):

zengin:
closing based on the decision from daily sync

Issue #1129, opened 2022-06-27
Pipeline not updating machine to latest powershell sdk version (1.10.0)


timayabi2020:
The powershell script defined in the powershell-execution-tests-template.yml file is not able to update to the latest powershell sdk version. It only installs the module if the command ```Get-Module -Name Microsoft.Graph -ListAvailable``` returns an empty list.
If the machine has an old version, powershell test runs will use that version.
<img width="407" alt="image" src="https://user-images.githubusercontent.com/10947120/175944135-65cd916d-96ba-4a14-973d-41a1a5bac695.png">

PR #1130, opened 2022-06-27
Updated sdk download script

timayabi2020:
- fixes #
This PR fixes https://github.com/microsoftgraph/msgraph-sdk-raptor/issues/1129

- Update the script defined in powershell-execution-tests.yml file. The script looks for the powershell sdk version equivalent to what is on the if-statement, if it is not found the latest version is installed. Therefore tests won't have to run on an older version since the previous script only checks if there exists a powershell sdk regardless of the version.
The image below is an illustration of how the script will behave when executed.
image

### DIFF
--- a/azure-pipelines/powershell-templates/powershell-execution-tests-template.yml
+++ b/azure-pipelines/powershell-templates/powershell-execution-tests-template.yml
@@ -7,20 +7,13 @@ steps:
   inputs:
     targetType: 'inline'
     script: |
-        $module = Get-Module -Name Microsoft.Graph -ListAvailable
-        if($module.Count -gt 0){
-          $module[0] | %{ 
-          if($_.Version -eq "1.10.0"){
+        $modules = Get-Module -Name Microsoft.Graph -ListAvailable
+        if ($modules.Count -gt 0 -and $modules[0].Version -eq "1.10.0") {
           echo "Latest version already installed"
-          }else{
+        }else {
           echo "Installing latest version"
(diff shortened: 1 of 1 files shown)


azure-pipelines/powershell-templates/powershell-execution-tests-template.yml
```diff
@@ -7,10 +7,22 @@ steps:
...
+          }else{
+          echo "Installing latest version"
+          Install-Module Microsoft.Graph -Repository PSGallery -Scope AllUsers -Force -AllowClobber
+          Import-Module Microsoft.Graph -Global -Force  
+          }}
+        }else{
+          echo "Installing latest version"
+          Install-Module Microsoft.Graph -Repository PSGallery -Scope AllUsers -Force -AllowClobber
+          Import-Module Microsoft.Graph -Global -Force
         }
```
zengin, line 24: `%` loops over all items in a list, `module[0]` is an item in the list, so we don't need a loop there. I think we can simplify is as follows:
```suggestion
        $modules = Get-Module -Name Microsoft.Graph -ListAvailable
        if ($modules.Count -gt 0 -and $modules[0].Version -eq "1.10.0") {
          echo "Latest version already installed"
        }
        else {
          echo "Installing latest version"
          Install-Module Microsoft.Graph -Repository PSGallery -Scope AllUsers -Force -AllowClobber
          Import-Module Microsoft.Graph -Global -Force  
        }
```

zengin: In addition, is there a way to check the latest version dynamically?

cc: @peombwa

Issue #1112, opened 2022-06-15
Update go sdk string template to include models, configs and other namespaces


fey101:
Related to PR https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/1021

PR #1141, opened 2022-07-05
Enhance go-template-to-include-models-and-configs

fey101:
Closes https://github.com/microsoftgraph/msgraph-sdk-raptor/issues/1112

Adds :
 - Logging as optional code only if there's a return value on the call to `graphClient`
 - import "logs" package in snippet file only if condition 1 above is true
 - import "time" package in snippet file only if `time.parse` appears in the compiled snippet.
 - optional models import if models is required in the snippet file
 - optional dynamic config import only if `graphconfig` appears in the compiled snippet.


TestsCommon/GoTestRunner.cs
```diff
@@ -141,6 +151,38 @@ private static string ReplaceHyphensWithUnderscores(string stringContent)
...
+            configParentString = configParentStringMatch.Groups[0].Value;
+        }
+        else
+        {
+            Assert.Fail("Regex {0}, against code {1} Failed", ConfigParentRegex, codeContentString);
+        }
+        var configSegments = configParentString.Split(".");
+        var configSegmentsRequired = configSegments[1..^1];
+        var requiredConfigPath = string.Join("/",configSegmentsRequired.Select(segment => segment.Split("(").First()))
+            .Replace("ById", "/Item")
```
rkodev, line 178: example `PlacesById("place_id")` will result in `Places/Item("place_id")`  for this statement

fey101: No. I have split on the '(' and returned only the first element of the split in the statement `configSegmentsRequired.Select(segment => segment.Split("(")` , so that we are only left with 'places/item/' in the above scenario.

fey101:
> I wonder how much of that logic should in facts be in the snippets generation logic? This way you could not only parse them for raptor, but it'd guide people in the import section (which has been a complaint in a couple of issues on the Go repo) Thoughts?

@rkodev ? How possible is this in snippet generation logic given that you are the one who looked into these changes in devX.

rkodev:
@baywet @fey101 It would be significantly simpler to add the logic at the snippet generation phase i.e add all the import statements at that point, plus I believe displaying a snippet with all the relevant dependencies would improve the overall dev experience, simply copy and paste.

We however don't display the imports for any of the languages at the moment, should we consider adding this to all the other languages.

fey101:
Great suggestion. I was wondering what decisions initially guided the snippets not having the import statements. @baywet? But I agree, having import statements in snippet generation would simplify work for users and even for raptor. Ofcos, raptor would have to make adjustments to accomodate these changes for all other languages as well.

baywet:
That decision predates my time on the team. Maybe @MIchaelMainer has more context there?

MIchaelMainer:
I'm not familiar with the decision to not include import statements. I will point out that adding import statements would result in repetitive content for each snippet. The same could be said about showing how to instantiate the graph client.  I can see why not including the import statements is an issue. The boilerplate documentation under the generated snippets could somewhat mitigate this by linking to documentation that describes which imports are needed. As it is now, there is no clear communication on what import statement is needed. Same could be said of what package is needed for the snippet.

"For details about how to [add the SDK](https://docs.microsoft.com/en-us/graph/sdks/sdk-installation) to your project and [create an authProvider](https://docs.microsoft.com/en-us/graph/sdks/choose-authentication-providers) instance, see the [SDK documentation](https://docs.microsoft.com/en-us/graph/sdks/sdks-overview)."

Having the import statements in generated snippets will be a requirement if we move away from monolithic client libraries and create feature/functionality-based segmented client libraries.

PR #1156, opened 2022-07-13
Increase timeout for beta pipelines

fey101:
The beta pipelines are timing out without completing all tests owing to the bulkiness of the go-beta-sdk and the increased number of tests in beta compared to V1. 

To prevent timeout, am increasing pipeline timeout to 4hours.


MIchaelMainer:
How many minutes will it now take with this changed to 240 minutes? Do you have an idea about when 240 will not be enough based on the size of the beta metadata?

Are you waiting for these builds to complete to perform a subsequent step in your work? If so, how often does it affect your inner loop? 

I ask as we may want to use more capable VMs.

fey101:
> How many minutes will it now take with this changed to 240 minutes? Do you have an idea about when 240 will not be enough based on the size of the beta metadata?
> 
> Are you waiting for these builds to complete to perform a subsequent step in your work? If so, how often does it affect your inner loop?
> 
> I ask as we may want to use more capable VMs.

Most times it takes approximately 1hr 50sth seconds and runs successfully. But sometimes it goes beyond 2 hours and is cancelled prematurely so it's hard to tell especially given it takes a much longer time locally owing to limited resources. I don't anticipate any time where it should go beyond the 4 hours. Probably just slightly over 2 hours but subject to confirmation with a couple more runs.

zengin (approved):

PR #1207, opened 2022-08-11
Fix stderr informational error  blocking GoBetaTests as a result of a…

fey101:
… new start to the message

Go Beta Tests failing as a result of following informational message being logged into the stderr: "go: added github.com/microsoftgraph/msgraph-sdk-go v0.33.0"

by replacing the special casing from `go: downloading` to just `"go: `, this includes all go sdk related informational logs in the special case for not logging error.


TestsCommon/GoTestRunner.cs
```diff
@@ -221,7 +221,7 @@ public static async Task DownloadRequiredDependencies()
         // special case empty lines or lines starting with "go: downloading" since they are not errors
         // Workaround to be removed once go sdk fixes this.
         if (stderr.Split(Environment.NewLine)
-            .Any(line => !string.IsNullOrEmpty(line.Trim()) && !line.StartsWith("go: downloading", StringComparison.InvariantCulture)))
+            .Any(line => !string.IsNullOrEmpty(line.Trim()) && !line.StartsWith("go: ", StringComparison.InvariantCulture)))
```
baywet, line 224:
```suggestion
            .Any(static line => !string.IsNullOrEmptyOrWhiteSpace(line) && !line.StartsWith("go: ", StringComparison.InvariantCulture)))
```

baywet (changes requested):

baywet (approved):

